PR #3, opened 2015-01-24
Trying to make this usable as a library

zerebubuth:
With the changes below, it becomes possible to use timely as a library, for example in [this little program](https://gist.github.com/zerebubuth/00e022087a3fc3ede855) based on an extract of timely's `src/main.rs`. However, I'm not entirely sure that it's working properly, as the example seems to loop endlessly?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,0 +1,11 @@
+#![feature(unsafe_destructor)]
+
+extern crate serialize;
+extern crate core;
+
+pub mod networking;
+pub mod serialization;
+pub mod progress;
+pub mod example;
+pub mod communication;
+
--- a/src/progress/broadcast.rs
+++ b/src/progress/broadcast.rs
@@ -4,7 +4,7 @@ use std::sync::mpsc::{Sender, Receiver};
 use progress::Timestamp;
 use communication::ChannelAllocator;
 
-type ProgressVec<T> = Vec<(u64, u64, T, i64)>;  // (scope, [in/out]port, timestamp, delta)
+pub type ProgressVec<T> = Vec<(u64, u64, T, i64)>;  // (scope, [in/out]port, timestamp, delta)
 
 // a mechanism for broadcasting progress information within a scope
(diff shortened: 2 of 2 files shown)


frankmcsherry:
I'll check it out. It is possible that it is working correctly, in that the intended behavior of `pull_internal_progress` is to give each dataflow scope/vertex the opportunity to do some work, and then return `true`/`false` to indicate whether there remains more work to do not reflected in the progress information passed back (e.g. unprocessed side-effects). If the loop instead consults the result of the method and terminates when false, I think it should terminate without looping indefinitely (that would be the intended behavior, at least; system bugs notwithstanding).

frankmcsherry:
Edits (and example program) look totally sane. I tweaked the numbers in my current main.rs to confirm, but it should eventually terminate. The magic numbers are in the lines 70-71 in the example, where the `1000000` indicates the number of iterations to do before suppressing the circulating records. This number was chosen to give a sufficiently long profiling trace on my laptop under release build, but might be unpleasantly long for anyone else. Switching it to `1000` causes it to terminate in about 79ms for me.

frankmcsherry:
About to go and re-factor main.rs into tests and examples, swap Cargo.toml out, etc. Code may be weird until that lands. :)

zerebubuth:
Thanks!

I tried the example with `1000`, but `cargo run --release` was still running after 90s. I have no idea why, but I'll try again once you've landed the re-factor and perhaps the issue will become clearer.

frankmcsherry:
Should be some examples in `examples/`, and built with `cargo test` into `targets/examples/`. I have the numbers there as `1000`, and they appear (for me) to exit cleanly immediately. I pulled the code over from `main.rs` into a `queue.rs` and a `barrier.rs`, and it looks basically the same as your example (while loop rather than for loop).

Still some work for me to do cleaning things up (it seems `#[bench]` code in examples isn't used, or I don't know how to drive it yet; test builds without optimizations (seemed fine for examples, but could make for sad first numbers); other things I haven't noticed yet. :)

PR #4, opened 2015-04-12
Random Cleanup

gereeter:
This is a mishmash of various changes that I was making to help understand the codebase that I think improve things. I can separate them into separate PRs if you'd prefer (since the changes are largely unrelated to one another).

### DIFF
--- a/src/communication/exchange.rs
+++ b/src/communication/exchange.rs
@@ -19,12 +19,12 @@ pub trait ParallelizationContract<T: Timestamp, D: Data> {
 // direct connection
 pub struct Pipeline;
 impl<T: Timestamp, D: Data> ParallelizationContract<T, D> for Pipeline {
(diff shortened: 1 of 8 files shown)


frankmcsherry:
These look great, thanks! I'm going to take a bit to digest and maybe ask a question or two to make sure I grok the implications, though so far each time I've thought "hey wait ..." it ends up being "oh, smart".

frankmcsherry:
Well, this is very neat, and actually pretty inspiring. The use of references for subgraphs makes me think it might be possible to do the same with graphs (and communicators) all the way down, since they are supposed to be effectively enclosing scopes.

Thanks much, and let me know when I am next wherever you are so's I can buy a drink.

gereeter:
No problem, and thank you for the kind words.

PR #5, opened 2015-04-16
Made Scope implementations take slices instead of vectors when applicabl...

gereeter:
...e. Previously, the implementations were just expected to avoid deleting any inputs or changing the length of the vectors - this is now enforced by the types.


frankmcsherry:
Thanks! This makes much more sense (I have to admit, many of the interfaces were done as part of learning Rust, so if they seem deranged, they probably are. :))

PR #6, opened 2015-04-16
Stop cloning Communicators and use mutable references instead of Rc.

gereeter:
With this change, `Communicator`s are no longer wrapped in `Rc` and are no longer cloned. The `FeedbackObserver` needed to be changed to avoid referencing the lifetimes in `Graph`. The borrow fiddling that is now in `_create_subgraph` is a bit unfortunate, but should be factor out-able.

### DIFF
--- a/examples/barrier.rs
+++ b/examples/barrier.rs
@@ -1,6 +1,7 @@
 #![feature(core)]
 #![feature(test)]
 #![feature(unsafe_destructor)]
+#![feature(scoped)]
 
 /* Based on src/main.rs from timely-dataflow by Frank McSherry,
 *
--- a/examples/distinct.rs
+++ b/examples/distinct.rs
@@ -107,22 +107,26 @@ fn _create_subgraph<'a, 'b, G, D>(graph: &'a RefCell<&'b mut G>,
                                                                           Stream<'a, 'b, G, D>)
 where 'b: 'a,
       G: Graph+'b,
-      G::Communicator: Clone,
       D: Data+Hash+Eq+Debug+Columnar {
     // build up a subgraph using the concatenated inputs/feedbacks
-    let comm = graph.borrow_mut().communicator().clone();
-    let mut subgraph = (graph.borrow_mut().new_subgraph::<u64>(), comm);
+    let mut graph_borrow = graph.borrow_mut();
 
-    let (sub_egress1, sub_egress2) = {
-        let subgraph_builder = subgraph.builder();//RefCell::new(&mut subgraph);
+    let (subscope, sub_egresses) = {
+        let mut subgraph = (graph_borrow.new_subgraph::<u64>(), graph_borrow.communicator());
 
-        (
-            source1.enter(&subgraph_builder).distinct().leave(graph),
-            source2.enter(&subgraph_builder).leave(graph)
-        )
+        let sub_egresses = {
+            let subgraph_builder = subgraph.builder();//RefCell::new(&mut subgraph);
+
+            (
+                source1.enter(&subgraph_builder).distinct().leave(graph),
+                source2.enter(&subgraph_builder).leave(graph)
+            )
+        };
+
+        (subgraph.0, sub_egresses)
     };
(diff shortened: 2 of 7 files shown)


frankmcsherry:
Neat!

As a related question: I was tempted to re-do the `Stream<'a, 'b, Graph, Data>` type to just use `&'a RefCell<Graph>` rather than `&'a RefCell<&'b Graph>`, in the interest of cutting the number of lifetime parameters down (since until associated lifetimes, people have to read/write them). This has the downside that rather than use `RefCell<&'b Graph>` as a guard and let it drop out of scope, you would need to call `RefCell<&'b Graph>::into_inner()` to get your graph back.

Any thoughts? At some level, I do like the idea this introduces of explicitly "finishing" work on a subgraph. I'm not sure I'm clear on what it gives up, but `&RefCell<&mut T>` and `&RefCell<T>` seem pretty similar.  Perhaps the former insulates more against someone cloning the `RefCell` and causing a surprising panic?

gereeter:
How about `impl<'a, G: Graph + 'a> Graph for &'a mut G`? This would remove the lifetime parameter from `Stream` while still avoiding `RefCell::into_inner`. Previously, I was also thinking that `impl<'a, G: Graph> Graph for &'a RefCell<G>` would be a good idea, but that doesn't work with `communicator` returning a mutable reference.

frankmcsherry:
That seems smart. I'll try this out and report back. :)

frankmcsherry:
I think there is a bit of a glitch with the communicator stuff. `cargo bench` panics with a `'RefCell<T> already borrowed'`, which makes sense because to snag the `&mut Communicator` we need to borrow the `RefCell<G>` for the lifetime of the subgraph. This makes any stream trying to `enter` the subgraph sweat, because they are sitting on the same `RefCell<G>` and are hoping to use it.

This is good though, because I had to think about the right organization! I think `Subgraph` should probably take a lifetime parameter `'parent` and a parent `G: Graph`, in the form of a `&'parent RefCell<G>`, so that rather than have to pair the subgraph with a `&Communicator` it can just provide one by borrowing up against its parent and implement `Graph` directly. At least, that is what I'm going to try to do. :D

frankmcsherry:
I pushed a few things that are along this path. Mostly they are like your pairing of `(Subgraph, Communicator)`, except that rather than hold an open `&mut Communicator` we have a `RefCell<&mut Graph>` which can snag communicator information. It was too painful to return a reference to the communicator from the graph, so I just added a method in to take a `FnOnce` closure to run with the communicator as an argument. It seems to work, if a bit ...

The check-in is also really close  to looking a lot prettier by implementing `Drop` for the (Subgraph, RefCell<&mut Graph>) pair (Rust currently ICEs on it, but feels better if the logic moves to an explicit `seal` method). With a `Drop` impl, they can/could add the subgraph to the parent scope, and code that initially looks like
```rust 
fn blah( ... ) {
    let mut subgraph = SubgraphBuilder::<_, u64>::new(graph);
    let result = {
        let subgraph_builder = subgraph.builder();

        (
            source1.enter(&subgraph_builder).distinct().leave(),
            source2.enter(&subgraph_builder).leave()
        )
    };
    subgraph.seal();

    result
}
```
turns in to (where the first two lines might be collapsable with nicer ergo):
```rust
fn blah( ... ) {
    let mut subgraph = SubgraphBuilder::<_, u64>::new(graph);
    let subgraph_builder = subgraph.builder();
    (
        source1.enter(&subgraph_builder).distinct().leave(),
        source2.enter(&subgraph_builder).leave()
    )
}
```

gereeter:
Whoops! I'm not sure how I missed that borrowing issue.

Those changes look really nice, and I like the fact that you no longer have to pass a graph to `leave`. `with_communicator` is annoying, but I don't see any better solution than what you've done.

It occurs to me that the wrapping of parent graphs might eventually cause performance problems, as each layer of nesting adds another layer of indirection. I'm not sure how to get around this, though.

PR #7, opened 2015-04-17
Make .seal take the nested builder by value and stop wrapping the Subgra...

gereeter:
...ph in Option.

This cleans things up a bit and makes it a compile-time enforced rule that you can only seal a subgraph once.

### DIFF
--- a/src/progress/nested/builder.rs
+++ b/src/progress/nested/builder.rs
@@ -25,7 +25,7 @@ use progress::nested::*;
 // TODO : so that any .borrow_mut().method() would avoid runtime panics if we did not do the above.
 
 pub struct Builder<'a, G: Graph+'a, TInner: Timestamp> {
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Looks good! I think that stuff was in there in the hopes that the `Drop` impl ICE would go away, and then all that `seal` logic could live in `drop`, which (annoyingly) can't move out of `&mut self`. But if that happens, hooray and should be easy to swing things around some other way.

gereeter:
Well, you can do the same thing in `drop`, but it takes some unsafety (`ptr::read`). This would break if `Subgraph` ever got a `Drop` impl, but I didn't think that was planned.

frankmcsherry:
I put up an issue about a different approach to doing things, in case you have any thoughts. I figured doing things that way (issues, rather than random comments) was probably a better way of leaving breadcrumbs behind for other readers! :)

PR #9, opened 2015-04-18
Make ActiveStream and SubgraphBuilder take their graphs by value

gereeter:
I held back on the `impl GraphBuilder for &RefCell<GraphBuilder>`, but it should be possible (with a switch back to `with_comminicator`).

### DIFF
--- a/src/example_static/builder.rs
+++ b/src/example_static/builder.rs
@@ -17,7 +17,7 @@ pub trait GraphBuilder: Sized {
 
     fn communicator(&mut self) -> &mut Self::Communicator;
 
-    fn new_subgraph<'a, T: Timestamp>(&'a mut self) -> SubgraphBuilder<'a, Self, T> {
+    fn new_subgraph<'a, T: Timestamp>(&'a mut self) -> SubgraphBuilder<&'a mut Self, T> {
         let subscope = self.new_subscope();
         SubgraphBuilder {
             subgraph: subscope,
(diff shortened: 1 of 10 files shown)


frankmcsherry:
Looks good. GraphBuilder-by-value also makes the "linear-ness" of the whole thing a bit more clear. You really don't want to write methods that drop the graph on the floor. Sort of like how `partition` makes sure to return the `GraphBuilder` as well, which I didn't do at first. >.<

frankmcsherry:
One minor comment, which should be easy to fix: in trying out the `dataflow_join` code I noticed that `enable` always takes its `self` argument as `&mut self`. This sticks you with a `&mut G` which isn't always what you want (especially if you started with a `G` and are in some extension method). 

That could become `self`, I supposed it should (will do!) but then you'd have to write weird things like
```rust
(&mut builder).enable(&stream)
```
So, I'm going to uncomment `Stream::enable` which takes a `G` as input by value. It was commented out in an attempt to force people to think about the graph builder as doing the action, and maybe to save a few `&mut`, but I think we can try both of them now. I suspect we'll want to turn off one of them eventually.

The above just becomes:
```rust
stream.enable(builder) // or &mut builder, if you'd rather borrow
```
Maybe there is a better transitive verb to use here...

frankmcsherry:
Commit https://github.com/frankmcsherry/timely-dataflow/commit/e03779143a353a33ff4c8c84abefc38072030fb8 should do it!

gereeter:
I left those methods taking references simply because I thought that using `&mut` would still be the most common thing to do, and I thought that ergonomics would be hurt by taking the graphs by value. However, the by-value versions are strictly more general.

frankmcsherry:
That makes sense. The specific case I had was coming in with an `ActiveStream<G, D>` and having been given a bundle of "actions on `ActiveStream<G, D>`". They got all sorts of angry that I now had a `ActiveStream<&mut G, D>` in some of the places and not others. :)

PR #12, opened 2015-04-22
Move Graph and SubgraphBuilder into example/

gereeter:
These are really specific to the non-static method of building scopes.


frankmcsherry:
Good idea! 

There must be a better name than "examples" for each of these. They were originally "examples" of how to use the progress tracking logic, but are clearly the more appealing way to use things. Perhaps "dataflow" or such, being an example dataflow abstraction built on pre-existing coordination and communication substrates. No clue...

As a side comment, I switched out

```
impl<'a, G: GraphBuilder+'a> GraphBuilder for &'a G { ... }
```
and replaced it with (courtesy some folks on IRC)
```
impl<'a, G: GraphBuilder> GraphBuilder for &'a G { ... }
```
because a) I couldn't understand the distinction, and b) it was causing the borrow checker to unify some lifetimes that it "shouldn't have". It got buried in some pile of commits, to my great shame. But, if it strikes you that there is an important difference between the two let me know. It breaks a bunch of things for me, but better to understand what is going on. :)

PR #13, opened 2015-09-17
fix to parameters in multi-process example with cargo run

utaal:
`cargo run` usage is `cargo run [options] [--] [<args>...]` and only parameters after `--` are passed through to the spawned process.


frankmcsherry:
Ah, thanks!

PR #14, opened 2015-12-26
Switch the argument order on `enter` and `enter_at`. These are now me…

gereeter:
…thods on a `Stream`. You mentioned that you might like this in frankmcsherry/differential-dataflow#2. This may want a version bump because it is a breaking change.


frankmcsherry:
Yar, totally a version bump! I think this isn't actually current on crates.io at the moment because of a dependence on a a non crates.io crate as a dev-dependency. I'll try and clean that up.

PR #17, opened 2016-02-03
Capability-based api for operators

utaal:
Hi Frank,
here's a (mostly) cleaned up branch for my attempt at providing a capability-based API to enforce the operator contract.
~~It still needs docstrings and possibly one final rebase to cleanup the history.~~

A couple of things to note:

* `InputHandle`, `OutputHandle` and `Capability` can only be constructed from within the "operators" module, this should make it impossible(?) for someone to mint a new `Capability` out of thin air or to `mem::replace` the `*Handle`s; moreover the field of the `*Handle`s are private, which should prevent the user from tampering with the underlying `Counter`s and `Buffer`s. (And yes, it seems like you cannot make `impl` member package-private, if they're `pub` they're always re-exported when exporting the corresponding trait/struct and if they aren't `pub` they're impl-local. From what I gathered from #rust yesterday night this is the idiomatic way of module-private apis, they're talking about adding module-level visibility modifiers but they're still a long ways off.)
* ~~My approach to `Notificator` is hybrid, it's not relied upon for progress tracking anymore but it still retains most of the original logic.~~ (I changed my mind on this as I realized my "optimization" was in fact quite the opposite - I rewrote it, but I need to take another look at it.)

~~Feel free to wait for me to add the docstrings before going through the code, I just wanted to let you know that I haven't fled to Hawaii or something :)~~


utaal:
Oh, I have a branch of differential-dataflow that compiles with the capabilities api, the diff is small: `7 files changed, 41 insertions(+), 42 deletions(-)` (mostly a few uses of `capability.delayed(new_time)` and some explicit calls to `capability.time()` to correctly infer `Timestamp` as the key in a few hashmaps).

frankmcsherry:
Sweet! I'll take a peek tomorrow.

src/dataflow/operators/notificator.rs
```diff
@@ -0,0 +1,136 @@
...
+                        *count += 1;
+                        false
+                    },
+                    None => true,
+                }
+            };
+            if push_new {
+                self.available.push_back((newly_available, 1));
+            }
+            last = position;
```
gereeter: Should this be `last += position`?

frankmcsherry: I think so, right? `position` is relative to an iterator that starts from `last`, so to slide `last` forward we want `+=`. Let's see what Andrea thinks.

utaal: Yep, that's correct.

frankmcsherry:
Looks good to me, modulo possibly missing random gotchas (timely dataflow is hard). The design seems much simpler than what I had imagined, due I think to only worrying about `unary` and `binary` rather than things like `input`, `to_stream`, and `replay`. Probably better this way. :)

src/dataflow/operators/notificator.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub fn for_each<F: FnMut(Capability<T>, i64)>(&mut self, mut logic: F) {
+        while let Some((cap, count)) = self.next() {
+            ::logging::log(&::logging::GUARDED_PROGRESS, true);
+            logic(cap, count);
+            ::logging::log(&::logging::GUARDED_PROGRESS, false);
+        }
+    }
+
+    #[inline]
+    fn scan_for_available_notifications(&mut self) {
```
utaal: @frankmcsherry on second look, while this is a single scan through all the requested capabilities, its semantics are different from the original `notificator`, which would always return the available notifications in (some) topological order (w.r.t. the timestamp's partial order) while this returns the capabilities in the same order as the notifications were requested (`.notify_at`). It feels like there may be some code that depends on the original behaviour and, regardless, it may be less surprising to always deliver notifications in-order (or, more correcly, never deliver notifications out-of-order).

There's also a tricky semantic difference when a new notification is requested while iterating through the available ones: the original notificator would still deliver it in-order (because the only notifications that can become available are the ones in the maximal(minimal) antichain) while this technique would possibly deliver it out-of-order if another notification with a larger timestamp was already collected in the `available` `VecDeque`.

What do you think? Should we strive to retain the ordering?

frankmcsherry: Hum. Good point. I think this is one point where the intended guarantees were "no promises", but we've certainly written code that depends on the ordering.

If you want to be lazy, the notificator I wrote when trying this just had `pending` and `frontier`; no `available`. It just scanned the list on each call to `next`, which I feel slightly---but not too---bad about.

I think in the glorious future we should be getting a function from capabilities to iterators, and perhaps then it matters less what order they get called in (at least, it gets slightly harder to write code that cares).

I think the advantage to doing everything in order is that operators capable of compacting state know that when they get a notification they can compact times less than that of the notification; neither messages nor notifications will be arriving. Doing notifications out-of-order means they would have to track a notification frontier too.

utaal: > If you want to be lazy, the notificator I wrote when trying this just had pending and frontier; no available. It just scanned the list on each call to next, which I feel slightly---but not too---bad about.

Were you somehow checking to ensure you're delivering one of the notifications in the minimal antichain or was it simply the first one you encountered that was below the frontier?

> I think the advantage to doing everything in order is that operators capable of compacting state know that when they get a notification they can compact times less than that of the notification; neither messages nor notifications will be arriving. Doing notifications out-of-order means they would have to track a notification frontier too.

Yup, I also talked with Zaheer (as a "user") and while he doesn't think he has written anything that relies on this:
* it certainly seems like that operators that keep inter-epoch state (sliding window and similar) need to have more complex logic if no guarantees are given on notification delivery order;
* we both believe that it may be better (as in: less surprising to newcomers) if the ordering was guaranteed;
I'll ask around in the meeting as well to know what people think the current guarantees are.

An example of operators that rely on ordering (albeit possibly not the most interesting one) is https://github.com/frankmcsherry/timely-dataflow/blob/master/src/dataflow/operators/queue.rs#L24

Considering that the changes in user-level code are minimal for the switch to capabilities, maybe we should just make the ordering guarantee explicit for this first iteration of the capability-based api to avoid subtly breaking a bunch of code? And then maybe (maybe) provide different notificators (in the scenario with completely user-level notificators) depending on the guarantees that the user is looking for?
My first re-implementation preserved the original semantics by using the non-capability-aware logic and stashing capabilities on the side: https://github.com/utaal/timely-dataflow/blob/capabilities-with-ordered-notifications/src/dataflow/operators/notificator.rs (and, yes, `LazyKeyedSet` needs a new name).

frankmcsherry: I cannot confirm that I wrote the code correctly. :)

But, I think if you scan `pending` for the index of an available capability and update the index if you find a capability less than your current one, that should guarantee that they are delivered not-out-of-order. Right, maybe? :)

utaal: > But, I think if you scan pending for the index of an available capability and update the index if you find a capability less than your current one, that should guarantee that they are delivered not-out-of-order. Right, maybe?

Yup, sorry, I wasn't sure whether you meant that your code was preserving the guarantee (by scanning the entire list) or was just grabbing the first available notification.
I've came up with a (maybe) similar approach, https://github.com/utaal/timely-dataflow/blob/ced70769e866d5d756e7da0657757aa24f5c03e0/src/dataflow/operators/notificator.rs
Can we go with something like that?

> Edit: Also, I think `queue` works without order constraints on notifications. It keeps each queue differentiated by time. What am I missing?

I may have misunderstood what `queue` is supposed to do, but if it can be used to ensure that data is sent in-order to its downstream operators then I think out-of-order delivery of notifications would break it.

frankmcsherry: That might work! I don't read functional code; lemme try google translate on it. Would your life be easier if notifications didn't have a multiplicity, and we just de-dup on insert?

Whilst yoga-ing I was supposed to be thinking of birds and shit, but instead was thinking about how in the fullness of time we probably want the notificator to be throughput optimized (differential wants multiple concurrent notifications, and then to sort work by key rather than by time). This probably means "serve all at once". Just getting that out there.

Queue was *literally* the first timely operator I ever wrote (I think)! It was just supposed to batch things up, not deliver them in any particular order. Actually, it was just supposed to stop an infinite delivery loop. Oops.

utaal:
Ahaha, in my defense I wrote it imperative first but it made the borrow checker mad; then I realised I was actually writing a fold and I couldn't resist :)

Btw, just realised it's incorrect (it's not cleaning up dupes, I'm out to dinner, I'll fix it later).

And yes, de-duping on insert would make it boil down to a .min_by(|cap| cap.time()), I think.

src/dataflow/operators/notificator.rs
utaal: Ahaha, in my defense I wrote it imperative first but it made the borrow checker mad; then I realised I was actually writing a fold and I couldn't resist :)

Btw, just realised it's incorrect (it's not cleaning up dupes, I'm out to dinner, I'll fix it later).

And yes, de-duping on insert would make it boil down to a .min_by(|cap| cap.time()), I think.

utaal: Ok, so, I think I may have come up with something which may be a reasonable compromise between minimizing the number of scans and being somewhat readable rust.
There's no (little) functional-style code, it should guarantee in-order delivery and it's relatively compact. I've stared at it for a while and I haven't found any flaws but I should probably write a test.

utaal:
Added an arguably decent test of `Notificator`'s guarantees; it passes!

frankmcsherry:
Ok, Andrea is on the hook for immediate breakage. :)

PR #18, opened 2016-02-06
Make CountMap more opaque

gereeter:
This is a breaking change, but it leaves more freedom to change the representation of `CountMap` later.


frankmcsherry:
Looks good! Thanks again. :D

PR #20, opened 2016-02-15
Fix example output

mre:
I've heard from your library through Mathieu Poumeyrol's blog post about [Rust and BigData](http://www.poumeyrol.fr/2016/02/15/Embrace-the-glow-cloud/).

To be honest I had a hard time following along since most of what timely-dataflow is doing is completely new to me. Fortunately the first example in the README was clear and concise. To my surprise the sample output did not match my expectations. So I ran the code on my local machine and got some different output which makes a lot more sense to me. I'm sure it's just a typo, but to safe others from the same confusion, I've decided to send you that little pull request.
Thanks for the cool project, even though it's a bit above my head right now. :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,16 +49,16 @@ You can run this example from the root directory of the `timely-dataflow` reposi
 ```
 % cargo run --example hello
 Running `target/debug/examples/hello`
-hello 0
-hello 1
-hello 2
-hello 3
-hello 4
-hello 5
-hello 6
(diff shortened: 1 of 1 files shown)


frankmcsherry:
How do I screw things like this up? :)

Thank you very much!

frankmcsherry:
... probably in the same way I screw up "clicking on buttons".

PR #22, opened 2016-02-16
Proxy indicator for "useful work" being performed by an operator in the ScheduleEvent logs

utaal:
This seems to be of interest for some of the analyses we're looking at. It infers whether and operator had a chance to make progress by looking at the progress update buffers. I made sure to make rust/llvm strip the code that goes through the `CountMap`s when logging is disabled.
Let me know if you think this is interesting enough to land on master or we should maintain it in a separate branch internally.

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -190,14 +190,45 @@ pub struct MessagesEvent {
 
 unsafe_abomonate!(MessagesEvent);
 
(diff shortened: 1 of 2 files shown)


frankmcsherry:
Looks good. We'll probably want to eventually get to only logging the non-empty events, so that we don't spam people listening to the logs, but that can totally wait until we better understand exactly what to write down and when.

PR #25, opened 2016-03-18
Ad Stats Example

bfrog:
Not sure if you'd be interested in adding this as an example or not, maybe with some tweaks, but whatever, did it for myself to experiment and was cool. Thought I'd share

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,3 +30,4 @@ time="0.1.34"
 [dev-dependencies]
 getopts="0.2.14"
 rand="0.3.13"
+magnetic="1.0.1"
--- a/examples/adstats.rs
+++ b/examples/adstats.rs
@@ -1,0 +1,266 @@
+extern crate time;
+extern crate timely;
+#[macro_use]
+extern crate abomonation;
+extern crate rand;
+extern crate magnetic;
+
+use std::time::Duration;
+use std::sync::Arc;
+use std::thread;
+use std::collections::{HashMap, HashSet};
+use rand::Rng;
+
+use timely::dataflow::*;
+use timely::dataflow::operators::*;
+use abomonation::{encode, decode, Abomonation};
+use magnetic::mpmc::mpmc_queue;
+use magnetic::buffer::dynamic::DynamicBufferP2;
+use magnetic::{Producer, Consumer};
+
+/// Ad event types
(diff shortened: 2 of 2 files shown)


bfrog:
And I suddenly remember this only works on nightly right now, I'll redact it as it is now

frankmcsherry:
Thanks for the proposal. While examples are **great**, I'm a bit hesitant to build up the `dev-dependencies` section too much. At least for dependencies I don't yet understand (e.g. if something breaks, can I fix it).

If it ends up in a stable-friendly form let me know, and we can try and figure out how to include it.

bfrog:
Right, the key thing I was sort of thinking of here was a live computation that takes input from set of listener threads/workers (http, rpc, whatever) and feeds them into timely. The simplest solution seemed to be a mpmc queue which magnetic provides in a lock-free implementation. There's some other libraries that at least work on stable with the same lock free ideas (crossbeam) that I might try out.

bfrog:
So I was using magnetic here for its lock-free mpmc queue, crossbeam which works on stable also seems to have that functionality, I'll try that out and reopen

frankmcsherry:
Thinking this out a bit more, one of the things that timely doesn't do at the moment (rather, the `timely_communication` crate doesn't do) is allow your main thread to do anything while the computation is running. A call to `execute` blocks the main thread until all the work is done. 

Thinking about your example, it might be nice to be able to hand out a MPMC queue, or even a SPMC queue, to each of the workers, and have the main thread (or some other threads) populate it as the computation runs.

For this to work, this hunk of code 

https://github.com/frankmcsherry/timely-dataflow/blob/master/communication/src/initialize.rs#L116-L141

needs to return something like the guards, rather than blocking until they return. For example, if it just returned all the guards, that would sort of work (though, it wouldn't be very RAII user friendly). Maybe I should just wrap them in some sort of struct whose `Drop` impl joins them, and which could in principle support things like cancellation in the future.

frankmcsherry:
The most [recent commit](https://github.com/frankmcsherry/timely-dataflow/commit/e4456b2abcba2ae8748ce7adb95354a992d0b183) implements something like this, so that with a *PMC queue, it should be possible/natural for workers to read from a queue shared with (and populated by) the main thread.

bfrog:
Nice! I'll check it out

Issue #23, opened 2016-03-14
Broadcast data to all workers


sga001:
Is there an easy way to construct a broadcast operator (similar to exchange) that sends (i.e., copies) all stream data to all workers?

Right now I'm doing the following which sort of works, but I wonder if this is the right way of doing this.

```
root.scoped(|builder| {
  let index = builder.index();
  let peers = builder.peers();
  let (input, stream) = builder.new_input();
  let mut streams = vec![];

  for i in 0..peers {
      streams.push(stream.exchange(move |_| i as u64));
  }

builder.concatenate(streams).inspect(|x| println!("{}, {:?}", index, x));
```

Ideally, one could build something like: 
`
stream.broadcast().inspect(|x| println!("{}, "{:?}", index, x);
`
My guess is that this can be done provided one implements a new ParallelizationContract. I think the only difference between Exchange and this new Broadcast pact would be the implementation of the Push trait in pushers::exchange::Exchange no?

Any thoughts?

frankmcsherry:
This is a good issue to have open. The slightly complicated thing for a `ParallelizationContract` is that it isn't bright enough (without some help) to know how many copies of a record it would make, which is important information for the progress tracking logic (which needs to know how many records are at large on each channel).

In retrospect, this shouldn't be a problem because the channel has some number of targets, and makes some number of copies of each record, but something blocked it when I tried. Let me look into this this evening,  and either reconstruct what the issue was (and report back) or pop it in for you.

If it helps, what I do, which is less good than what the channel would do, is a `flat_map` that puts an index on records and then does an exchange, so you can broadcast even within a computation, rather than just at the inputs. But, less good for sure.

sga001:
Cool. Ideally one could implement something like Exchange but passing in a function that returns a collection of u64 rather than a single identifier. That way one can implement both broadcast and multicast with the same channel code.

frankmcsherry:
I looked into this, and it isn't all so horrible.

The thing I couldn't do before when I tried was just add a new implementor of `Push`, as a drop-in replacement for the existing `Exchange` implementor. This is because the progress tracking math is agnostic to where records end up, and so exists outside the routing decisions.

[...]

PR #26, opened 2016-03-24
Broadcast operator

utaal:
This is an attempt at providing a `.broadcast()` operator as discussed in #23 .

The internals create a new logical channel (`Target` port) for each destination worker and the message is broadcasted to all of them (by connecting each logical channel to the incoming `Stream`); having different channels (target ports) enables tracking progress for each of them separately.


sga001:
Thanks! this is exactly what I was looking for (but didn't quite know how to do). I'm curious to see what Frank thinks.

frankmcsherry:
I'm almost there! About to land at a desk in an hour and start the review! Really sorry! :).

frankmcsherry:
Ok, I totally understand this now. It took me much longer than I thought, though. :)

I think going forward with this is a good idea, to kick the tires. There is at least one optimization that will be important for a "great" implementation: only sending the data once per process. This seems like more of a pain with the current exposed API surface, though, so let's not get stressed yet.

Issue #27, opened 2016-03-29
Generic aggregate operator


sga001:
I've been thinking about implementing a generic aggregate operator (e.g., a reduce in MR lingo), since I've found myself writing the same pattern over and over.

The way I currently do aggregation is through a combination of unary_notify + state (let me know
if this is not the way to go!). It usually looks something like this:

```
let state =  // some collection (perhaps a HashMap<Timestamp, Vec<T>>). 

stream.unary_notify(Pipeline, "aggregate", vec![], |input, output, notificator| {
  input.for_each(|time, data| {
    notificator.notify_at(time.clone());
    state[&time].append(data.drain(..).collect());
  });

  notificator.for_each(|time, _| {
    let aggregate = //perform aggregation logic for all values of state[time]
    output.session(&time).give(aggregate);
   // garbage collect state[&time]
  }
}
```

Sometimes it is possible to update ``state`` as you go instead of appending to a list and updating at the end. So it'll look like this:

```
let state =  // some collection (perhaps a HashMap<Timestamp, T>). 

stream.unary_notify(Pipeline, "aggregate", vec![], |input, output, notificator| {
  input.for_each(|time, data| {
    notificator.notify_at(time.clone());
    state[&time] = // aggregation_logic(state[&time], data)
  });

  notificator.for_each(|time, _| {
    output.session(&time).give(state[&time]);
    // garbage collect state[&time]
  }
}
```

I was wondering if it would be easy to generalize these two cases so one can just do:

```
stream.aggregate(|x: Vec<T>| { aggregation logic that acts on x })  
```

```
stream.aggregate_it(|x: T, y: T| { aggregation logic in terms of x and y } )
```

Does this seem like a good idea? I think that all that would need to be done is something very similar to the way unary_notify is implemented (in unary.rs), but extend the struct Operator to incorporate this extra state (in addition to writing the necessary boilerplate)?

Any thoughts? Apologies if this is already covered by differential_dataflow's ``group`` operator (I honestly couldn't make sense of it).

frankmcsherry:
It definitely isn't covered by `group`, no worries. :)

When I've done this before, I've written something that takes `(K,V)` pairs, and is generic in an aggregation type `S` (for state) and a function `F: Fn(&K, V, &mut S)`. The intent is that as you get new data, you can describe how to update the state. 

Perhaps if you can update in place your function is

    |k, v, s| { *s += v; }  // S is just i64, maybe

[...]

PR #28, opened 2016-03-30
making trait public

sga001:
The aggregate operator seems to work for all of my intended use cases (see #27). The only minor change is modifying the visibility of the trait.


frankmcsherry:
Ha, oops!  Thanks. :D

PR #30, opened 2016-04-04
Rough attempt at input that supports multiple open epochs (with capabilities!)

utaal:
This is most definitely "work-in-progress" but I thought it would be a good idea to put it up for comments to see if you think this approach is worth spending more time on.

Take a look at the example for how its usage looks like. It enables multiple epochs to be open at the same time with similar semantics as "keeping capabilities" in a `*_notify/*_stream` operator. I wanted to see how burdensome it would be to have to deal with capabilities for input and it doesn't seem to be as terrible as I feared. What do you think? The only thing I'm very much not happy with is the fact that one needs to flush manually, but I think there are a couple of ways to get rid of that. And the nested tuple to return the input handle alongside the capability looks a bit awkward.


src/dataflow/operators/unordered_input.rs
```diff
@@ -0,0 +1,121 @@
...
+
+impl<T:Timestamp> Operate<T> for UnorderedOperator<T> {
+    fn name(&self) -> String { "Input".to_owned() }
+    fn inputs(&self) -> usize { 0 }
+    fn outputs(&self) -> usize { 1 }
+
+    fn get_internal_summary(&mut self) -> (Vec<Vec<Antichain<<T as Timestamp>::Summary>>>,
+                                           Vec<CountMap<T>>) {
+        let mut map = CountMap::new();
+        map.update(&Default::default(), self.copies as i64 - 1);
```
frankmcsherry: I think this might be a copy/paste issue from unary or the like. The `- 1` in those operators is because there is already a spare copy of the time in the `CountMap`, and we want to end with `self.copies` number in there. I *think* we probably want to ditch the `- 1` here; what do you think?

frankmcsherry: Ah, ok I think I understand. The capability that is minted is meant to increment the count, but I'm not sure it happens in the right point. When this method returns to the progress tracking infrastructure, timely will look and say "hey, not so many capabilities", and might do something rash. For example, in the workers = 1 case, it might say "no capabilities!" and confuse some other folks.

frankmcsherry: I think at this point in the execution, `self.progress` is properly configured and has some contents that we just want to pull out and multiply by `self.copies`. Ideally `self.progress` doesn't contain super important positive counts (which I think it does at the moment, right?), but rather we extract them here to ensure that the progress tracker sees them at start-up.

utaal: Yea, I wasn't super-careful with this. I was paying more attention to the API at this stage (e.g. it may be cool to auto-flush an epoch once its capability is dropped, which can't be done right now). Let me know if you think this chunk of code would be, in general, a good idea and I'll try and fix the "race" with your feedback.

frankmcsherry: Ah sorry. Yes, I should have said: looks good otherwise! ;)

src/dataflow/operators/unordered_input.rs
```diff
@@ -0,0 +1,123 @@
...
+    fn get_internal_summary(&mut self) -> (Vec<Vec<Antichain<<T as Timestamp>::Summary>>>,
+                                           Vec<CountMap<T>>) {
+        let mut internal = CountMap::new();
+        // augment the counts for each reserved capability.
+        for &(ref time, count) in self.internal.borrow().iter() {
+            internal.update(time, count * (self.peers as i64 - 1));
+        }
+
+        // drain the changes to empty out, and complete the counts for internal.
+        self.internal.borrow_mut().drain_into(&mut internal);
```
frankmcsherry: Totally minor (I think) but, couldn't this be combined with the iteration above as 

```rust
for (time, count) in self.internal.borrow_mut().drain(..) {
    internal.update(time, count * self.peers as i64);
}
```

frankmcsherry: Ah, maybe `CountMap` doesn't have a `drain(..)` method. That would be a good reason.

src/dataflow/operators/unordered_input.rs
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    pub fn session<'b>(&'b mut self, cap: &Capability<G::Timestamp>) -> Session<'b, G::Timestamp, D, PushCounter<G::Timestamp, D, Tee<G::Timestamp, D>>> {
+        self.buffer.session(cap)
+    }
+
+    pub fn flush(&mut self) {
+        self.buffer.cease();
+    }
+}
```
frankmcsherry: It seems like `UnorderedHandle` has (at the moment) just two interesting methods: `session` and `flush`, which just call through to `PushBuffer`'s `session` and `cease` methods. Would it reduce the cognitive overhead at all to remove this wrapper and just return a `PushBuffer`? Or does the wrapper add something else (e.g. a much shorter name to type out :)), now or in the future (e.g. the possible `Drop` impl)?

Alternately, should whatever pleasant things this type does be integrated into `PushBuffer`?

PR #31, opened 2016-04-21
fixed some clippy warnings

llogiq:
This just fixes some style nits and is probably somewhat opinionated. Even so, I hope it'll be useful.

### DIFF
--- a/src/progress/nested/summary.rs
+++ b/src/progress/nested/summary.rs
@@ -64,8 +64,8 @@ impl<SOuter: Display, SInner: Display> Display for Summary<SOuter, SInner> {
     #[inline]
     fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
         match *self {
-            Local(ref s) => f.write_str(&format!("Local({})", s)),
-            Outer(ref s, ref t) => f.write_str(&format!("Outer({}, {})", s, t))
+            Local(ref s) => write!(f, "Local({})", s),
+            Outer(ref s, ref t) => write!(f, "Outer({}, {})", s, t)
         }
     }
 }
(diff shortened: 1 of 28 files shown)


src/progress/nested/summary.rs
```diff
@@ -63,9 +63,9 @@ where TOuter: Timestamp,
 impl<SOuter: Display, SInner: Display> Display for Summary<SOuter, SInner> {
     #[inline]
     fn fmt(&self, f: &mut Formatter) -> Result<(), Error> {
-        match self {
-            &Local(ref s) => f.write_str(&format!("Local({})", s)),
-            &Outer(ref s, ref t) => f.write_str(&format!("Outer({}, {})", s, t))
+        match *self {
+            Local(ref s) => f.write_str(&format!("Local({})", s)),
+            Outer(ref s, ref t) => f.write_str(&format!("Outer({}, {})", s, t))
```
gereeter: Can these just be calls to `write!(&mut f, /*...*/)`?

llogiq: I think they can even be calls to `write!(f, /* ... */)` (because `f` is already a mutable reference).

frankmcsherry:
Thanks! Most of these look good, but there are a few that I'm not sure of: 

1. The plan/hope is to eventually remove `Copy` from the `Timestamp` trait, and all of those `.clone()`s are in there to try and keep things clear about when we need additional copies of timestamps (they are probably not sufficient). Though, maybe the best thing is to have no clones and comment out the `+Copy` to see what is broken.

2. The notificator iterator was written using `while let Some(...) = ...` to keep just one idiom in operator logic. The input "iterator" is a streaming iterator, so it only admits the `while let Some(...)` form of iteration. 

But ... maybe these are silly reasons to not just take all of the diffs. I should also go and install clippy probably, right? ;)

PR #34, opened 2016-05-12
Opening a Session borrows the capability for as long as the session is open

utaal:
This prevents dropping the capability while it's still "in use", i.e.
data is being sent at the associated timestamp


frankmcsherry:
Good catch!

utaal:
Saw it while looking at how (and if) to make the input handle and the operator's output session similar.

Issue #36, opened 2016-05-30
Scopes should separate channel allocation and operator ownership


frankmcsherry:
The current scope design has one struct that manages both channel and identifier allocation, and operator ownership. Because of the former, the scopes get cloned all over the place so that folks can create channels and identifiers. Consequently, ownership gets cloned all over the place and we end up with reference cycles, and resources don't get released until operators are tombstoned (which is not guaranteed to happen). 

Obviously this is horrible if you rely on releasing resources (or have any standards at all).

The fix doesn't seem too complicated: scopes like `Root` and `Child` should clearly distinguish between the cloneable `Allocator` types they have, and the non-clonable operators they own. When we need to clone a scope, we should determine that we only need the `Allocator` component (otherwise: weird), and just clone that.

Specifically, where `Root` currently looks like

```rust
pub struct Root<A: Allocate> {
    allocator: Rc<RefCell<A>>,
    graph: Rc<RefCell<Vec<Box<Operate<RootTimestamp>>>>>,
    identifiers: Rc<RefCell<usize>>,
}
```

calls to allocate channels and identifiers currently resolve to the first and third fields. These two should be bundled up into something called `Allocator` or so, and made available. At the same time, it would be helpful to remove the `Clone` implementations from scopes, as well as their `Rc` wrappers around owned operators, to clarify the single ownership.

frankmcsherry:
Also, I'm on this one, so no ninja pull requests fixing it for me. :)

kali:
oh good. I half tried doing it at some point a few months ago but never got to the end of it.

frankmcsherry:
More investigation reveals that the problem wasn't exactly cycles in scope ownership, which can't easily happen (the `Child` scopes own clones of their parents, but the parents do not naturally point at `Child` children).

Rather, the problem seemed to resolve to a scope getting captured by the closure of an operator, which is another problem altogether. The current fix is meant to be minimally invasive, and amount to extracting `Subgraph` from it's `Rc<RefCell<_>>` container before adding it as an operator. This ensures that the subgraph is uniquely owned. The downside is that it happens at graph construction time, rather than compile time.

[...]

PR #37, opened 2016-06-01
Statically enforce that references to child scopes don't escape

gereeter:
cc #36

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,8 +13,8 @@ repository = "https://github.com/frankmcsherry/timely-dataflow.git"
 keywords = ["timely", "dataflow"]
 license = "MIT"
 
-#[dependencies.timely_communication]
-#path="./communication"
+[dependencies.timely_communication]
+path="./communication"
 
 [features]
 default=[]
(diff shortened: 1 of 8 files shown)


gereeter:
This "provide[s] static guarantees that references to scopes do not escape the scope construction" as you mentioned you wanted in #36, but to avoid ergonomic problems does not change the API other than adding some lifetime parameters to `Child` - not a single line of example code was changed.

Also, sorry, I realize that you said no ninja pull requests, but this seemed to be simple enough that it was easier to write a PR than explain my reasoning, and it was different from the larger, more complicated problem of separating `Scopes` and `Allocator`s.

frankmcsherry:
No, this is great. The "no ninja PRs" was more meant to prevent disappointment if someone started and then found out that I did it.

I've looked at the code and it seems sane. I think (roughly a year ago) when we tried this last the horrible nightmare of lifetimes was due to making the other `Child` field (`parent`) be a reference, but this lifetime seems pretty painless...

frankmcsherry:
Yeah, this is very nice. Thank you very much! I'll be sure to publicly lament the next thing I can't figure out! :D

PR #38, opened 2016-06-05
removed time crate; converted times to Instant, Duration

frankmcsherry:
Edits to remove the `time` crate as a dependence, and consequently several of its dependencies. Impact seems limited only to the logging subsystem and the attractive output of two of the examples.

In addition to swapping in `std::time::Instant`, initialization creates one `Instant` and supplies it to all the loggers, who use it to generate timestamps rather than the shared static. This has the potential to reveal worker start-up skew in the reported times, which we could fix if we want by augmenting `Root::new()` with an `Instant` parameter for the logging, and which would be minted by `timely::execute()`.

Other folks are using the logging infrastructure, so I want to give them (and anyone, really) a chance to try out and speak up if this subtly breaks their work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,8 +25,15 @@ abomonation="^0.4.2"
 timely_sort="0.1.3"
 #timely_communication="^0.1.4"
 byteorder="0.4.2"
-time="0.1.34"
+#time="0.1.34"
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


frankmcsherry:
Conversations with others using the logging infrastructure suggests that this is not the right time to change how it works. It is probably worth doing this in the future, but not right now.

PR #39, opened 2016-06-21
Custom operator api that exposes the frontier directly

utaal:
In this PR, custom operator api that exposes the frontier directly
- it does not provide the notificator, that can be instantiated
  separately by the client;
- it supports requesting initial capabilities.

A notificator that can be instantiated by the user is included.

Names and namespaces are totally negotiable and there's some missing_doc in notificator.rs


examples/operator.rs
```diff
@@ -0,0 +1,100 @@
...
+                        }
+                    }
+                }).inspect(|x| println!("{:?}", x));
+        });
+    }).unwrap();
+
+    timely::execute(Configuration::Thread, |root| {
+        root.scoped(|scope| {
+            (0u64..10).to_stream(scope)
+                .unary(Pipeline, "example", |mut builder| {
```
utaal: If you're still interested in something like this, I think I have a new plan: instead of getting a builder (which complicates things), you request a capability (with its timestamp) and you get the actual cap in the closure. Something like:

```
.unary(Pipeline, "example", Some(RootTimestamp::new(12)), |mut cap| {
  // cap is now a Option<Capability> = Some(Capability for timestamp 12)
  // you then `move` it into the actual operator logic

  move |input, output| {
    if let Some(ref c) = cap.take() {
      // ...
```

The outer closure can still be used to instantiate a notificator.

```
         in1.unary_frontier(&in2, Pipeline, Pipeline, "example", None, |_| {
            let mut notificator = FrontierNotificator::new();
            move |(input1, frontier1), output| {
              for time in notificator.iter(&[frontier1]) {
```

This could (probably) be improved by handing the reference to the frontier/progress updates to notificator directly, instead of having to refer to it from inside the operator logic.

```
         in1.unary_frontier(&in2, Pipeline, Pipeline, "example", None, |_, frontier1| {
            let mut notificator = FrontierNotificator::new(frontier1);
            move |input1, output| {
              for time in notificator.iter() {
```

If this sounds reasonable, let me know and I can take a stab at it. But if doesn't line up with what you have in mind, feel free to close this PR.

frankmcsherry: Another option (other than requesting capabilities) is just to be given a capability for the default (least, I hope) time. You can then `delayed` it to whatever times you need, or just drop it if you don't want it. This is done at initialization time, so it shouldn't be terribly expensive.

You could imagine writing something like

```rust
stream.unary(exchange, "example", |cap| {}, |input, output| {
    // logic ..
});
```

where if you'd like an initial notification, you do something with `cap` in that closure. 

This has the annoying property that your notificator would be borrowed by both closures, and so you would need to `Rc<RefCell<_>>` it up. Perhaps the constructor approach you are suggesting is the tasteful way to get around that? In which case, we could simplify it by removing the "request" part and just always giving you a `Capability::new(Default::default())` or whatever the least time is. You would write semi-gross things like:

```rust
stream.unary(exchange, "example", |cap| { |input, output| {
    // logic ..
}});
```

which maybe isn't horrible? I haven't tried, but it would be cool to know if all the ownership works out ok, or if this ends up a pain.

Having written a few weird operators ("weird": not much use for traditional notificator), I think what I'd love to see is: 

1. Frontier for each input as part of `InputHandle`. This seems to make so much sense. For operators that don't request frontier information, we just keep it empty (or make it an `Option<&[Timestamp]>`? Seems like a pain)
2. `Notificator` demoted to user code; you should only get one if you mint it up and use it.
3. Perhaps "capability list" associated with outputs, dual to the frontiers for inputs? I'm usually just writing against one output and holding my capabilities separately; perhaps I should keep doing that.

How do these strike you?

utaal: 1. Are we talking about a computed frontier

utaal: That makes sense, I think I still kinda prefer the two nested closures: no need for Rc<RefCell<_>> and you get a reasonable spot where you can initialise local state or a user-space notificator, i.e.:

```
long().stream().composition()
  .that().spans().multiple().lines()
  .unary(exchange, "example", |cap| {
    let stash = Vec::new();
    let notificator = FrontierNotificator::new();
    move |input, output| {
      // logic ..
    }
  });
```

I tried to do 1. (Frontier in `InputHandle`) and it worked out pretty well. I made the `Option` private and only exposed a "method" that calls `.expect()` so that the user doesn't need to unwrap manually and the computation would still blow up if someone tries to use the frontier on an input that doesn't support it. Code here https://github.com/frankmcsherry/timely-dataflow/pull/39/commits/c7f768af1cd42c223097d07159de07f97e516cfb

2. (user-level `Notificator`) is already a thing (take a look at the examples in this PR): does it match what you had in mind?

I'm not sure I understand 3. Right now the capabilities are "internal"; how would capabilities associated with output(s) look like?

frankmcsherry: Thanks! Elbows deep in diff dataflow stuff at the moment, but will try and check out code today. Nested closures does sound better than alternative (and better explains actual ownership).

About 2.: Sorry, I should have looked harder. I assumed because it managed the frontiers it had to be blessed so that `push_external_progress` would know to have a chat with it and update the frontiers. Let me take a peek, but that was the only magical thing I thought needed to be removed.

About 3.: no clue. Sorry, that was a bit more speculative and maybe nonsense. I've often found myself with a big pile of `(data, time)` tuples, and want to track down the associated capabilities which I just keep in a list, and I was trying to think about some place sane to put them.

frankmcsherry: Code looks good!

Sorry for the silly question, but .. if the method is `unary_frontier`, and presumably there is a `unary_no_frontier` or something, couldn't/shouldn't we just have two different `InputHandle` types that each of their closures take? That way we structurally prohibit frontier-checking errors.

utaal: Tried this in `https://github.com/frankmcsherry/timely-dataflow/pull/39/commits/b53c69e4074e76c673ef61dcdff03615993dcc23`: as is it requires an additional "use" in files that define operators (to have `InputHandler.next()` and `.foreach()` in scope.

frankmcsherry: Is the trait helpful? I'm reading it trying to grok, and not sure if one couldn't just skip it and expose the two structs directly.

frankmcsherry: Read things. Correct me if I'm wrong, but it looks like the trait exists because these two have similar methods, but no code uses the trait for polymorphism. So, we could skip the trait and put the methods in the `impl` for each type, if we wanted to skip the `use` statements.

utaal: Yep, correct.

utaal: I've removed those traits, and removed the builder stuff. You can take a look at `example/operator.rs` [here](https://github.com/frankmcsherry/timely-dataflow/pull/39/commits/64479d5175f61369344d035eda8b6151d00c86be).

frankmcsherry:
Looks great! Could I ask you to do some documentation that TravisCI is complaining about? (timely is now `#[deny(missing_docs)]`).

utaal:
Yep, yep. I commented that out while hacking on it :p
Shall we just merge this in as a separate operator interface or re-writer
other operators in terms of this? (see my other email)

utaal:
I cleaned this up, added missing_docs, and added examples for the new operator interface and new user-level notificator. I think it may be ready to be merged.

frankmcsherry:
Nice! I'll start looking in to the user-level notificator stuff, and propose various bits of clean-up around them.

PR #44, opened 2017-02-13
Fix compiler warnings

gandro:
Fixes a bunch of compiler warnings. The first one might become a hard error in the future, the rest are unused imports and some stylistic warnings.

There are still a couple of warning issued in the doctests, but many of them serve illustrative purposes in the examples (i.e. unused result for `let output = ...;`), so I didn't change these.


frankmcsherry:
Thanks very much!

PR #54, opened 2017-03-29
Refactor timely-sort

gereeter:
A good chunk of the complexity in `timely-sort` relates to small details of memory management - specifically, managing a collection of reusable buffers and using batches of `Vec`s instead of a simple `Vec`. This pulls out much of that logic, which is shared between the different sorting algorithms.

In a number of cases, optimizations were inconsistently applied between the algorithms, so standardizing improves performance.

### DIFF
--- a/sort/src/batched_vec.rs
+++ b/sort/src/batched_vec.rs
@@ -84,12 +84,16 @@ impl<'a, T> BatchedVecRef<'a, T> {
     }
 }
 
+/// An efficiently-appendable collection of `T`s. Unlike `Vec<T>`, which stores all elements contiguously,
+/// `BatchedVec<T>` stores elements in a bunch of independent batches, avoiding the need to periodically
+/// copy lots of elements.
 pub struct BatchedVec<T> {
     tail: Vec<T>,
     batches: Vec<Vec<T>>
 }
 
 impl<T> BatchedVec<T> {
+    /// Create an empty `BatchedVec<T>`
     pub fn new() -> BatchedVec<T> {
         BatchedVec {
             tail: Vec::new(),
@@ -97,6 +101,10 @@ impl<T> BatchedVec<T> {
         }
     }
 
+    /// Actually access this `BatchedVec<T>`. Because of the `BatchedVecX256` optimization,
+    /// we don't want to implement all the methods on `&mut BatchedVec<T>`, since `BatchedVecX256`
+    /// has no way of getting those thin references. Instead, we have a custom "fat pointer" type
+    /// that all access is done through, generalizing `BatchedVec` and `BatchedVecX256`.
     pub fn ref_mut(&mut self) -> BatchedVecRef<T> {
         BatchedVecRef {
             tail: &mut self.tail,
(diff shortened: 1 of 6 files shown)


gereeter:
In addition, I noticed that all three sorting algorithms reimplement `per_cache_line` and `lines_per_page`. However, LSB sorting (no write combining) is the only algorithm that allows changing batch sizes from the default, MSB sorting has a different value for `lines_per_page`, and MSB sorting doesn't even use `per_cache_line` or `lines_per_page`, using the constant `1 << 10` instead. Presumably, these should all be basically the same thing, but I'm not sure what the correct choice is.

sort/src/batched_vec.rs
```diff
@@ -0,0 +1,150 @@
...
+        let len = self.tail.len();
+        ptr::copy_nonoverlapping(elements.as_ptr(), self.tail.as_mut_ptr().offset(len as isize), elements.len());
+        self.tail.set_len(len + elements.len());
+    }
+
+    #[inline(always)]
+    pub fn finish_into(&mut self, target: &mut Vec<Vec<T>>, stash: &mut Stash<T>) {
+        target.extend(self.batches.drain(..));
+        if !self.tail.is_empty() {
+            target.push(mem::replace(&mut self.tail, stash.get()));
```
frankmcsherry: I think in this sort of situation, we might want to avoid `stash.get()` and rather leave behind an unallocated `Vec::new()`, under the reasoning that while we are not using the sorter it shouldn't sit on allocations (other than those in `stash`, which we can manage with `rebalance`). Instead, we can wait until we receive elements the next time around, which should grab an allocated buffer when it sees a capacity zero buffer.

sort/src/lsb_swc.rs
```diff
@@ -109,75 +109,57 @@ impl<T> RadixShuffler<T> {
...
-                }
-                let front_len = self.fronts.get_unchecked(byte).len();
-                ::std::ptr::copy_nonoverlapping(self.staged.as_ptr().offset(per_cache_line!(T) as isize * byte as isize),
-                                                self.fronts[byte].as_mut_ptr().offset(front_len as isize),
-                                                per_cache_line!(T));
+            if *self.counts.get_unchecked(byte as usize) as usize == per_cache_line!(T) {
+                let staged_elements = slice::from_raw_parts(
+                    self.staged.as_ptr().offset(per_cache_line!(T) as isize * byte as isize),
+                    per_cache_line!(T)
+                );
```
frankmcsherry: This is much better than what I did. :)

frankmcsherry:
This looks cool, thank you! I made one comment about whether `finish_into` leaves behind allocated data or not (I think it should not, because we risk forcing an allocation (if `stash` is empty) and in either case we just sit on the memory while not using the sorter.

Other than that, looks good to me. Let me know if you agree / disagree with this issue!

Thanks again. :D

sort/src/batched_vec.rs
gereeter: Ok, I can do that. If I remember right, this was an instance where the different sorters had different behaviour, and the benchmarks didn't really show a difference. I'd say that you're right - possibly saving memory is the better thing to do.

frankmcsherry: Cool, thanks! If you'd rather not, I can totally swap it in afterwards. I can totally believe that it doesn't show on the benchmarks, but .. I end up using this in lots of weird places and sometimes you just sort 10 elements and avoiding 256 allocations helps. ;)

gereeter:
Ok, I fixed up the allocation in `finish`. As expected, any benchmark changes seem small enough to be noise.

frankmcsherry:
Sweet! Just waiting for the Travis CI build to clear and a-merging we will go!

frankmcsherry:
Thanks very much! I'm still not clear on the right way to have drinks delivered to you, but I appreciate the help! :D

frankmcsherry:
I just shipped these changes to crates.io as v0.1.4: https://crates.io/crates/timely_sort

PR #57, opened 2017-04-19
Implement TimeExchange PACT and extend Exchange operator

antiguru:
Implement exchanging by time and data vs. by data only. This is useful to have permuting worker assignments for skewed data.

### DIFF
--- a/src/dataflow/channels/pact.rs
+++ b/src/dataflow/channels/pact.rs
@@ -28,7 +28,7 @@ impl<T: 'static, D: 'static> ParallelizationContract<T, D> for Pipeline {
     }
 }
 
-/// An exchange between multiple observers
+/// An exchange between multiple observers by data
 pub struct Exchange<D, F: Fn(&D)->u64+'static> { hash_func: F, phantom: PhantomData<D>, }
 impl<D, F: Fn(&D)->u64> Exchange<D, F> {
     /// Allocates a new `Exchange` pact from a distribution function.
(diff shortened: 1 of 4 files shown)


src/dataflow/channels/pact.rs
```diff
@@ -47,6 +47,27 @@ impl<T: Eq+Data+Abomonation, D: Data+Abomonation, F: Fn(&D)->u64+'static> Parall
     fn connect<A: Allocate>(self, allocator: &mut A, identifier: usize) -> (Box<Push<(T, Content<D>)>>, Box<Pull<(T, Content<D>)>>) {
         let (senders, receiver) = allocator.allocate::<Message<T, D>>();
         let senders = senders.into_iter().enumerate().map(|(i,x)| Pusher::new(x, allocator.index(), i, identifier)).collect::<Vec<_>>();
+        (Box::new(ExchangeObserver::new(senders, move |_, d| (self.hash_func)(d))), Box::new(Puller::new(receiver, allocator.index(), identifier)))
+    }
+}
+
+/// An exchange between multiple observers
+pub struct TimeExchange<D, T, F: Fn(&T, &D)->u64+'static> { hash_func: F, phantom: PhantomData<D>, phantom_t: PhantomData<T>, }
```
frankmcsherry: If you like (it's minor) you can totally use `PhantomData<(D, T)>`. I think that is how folks do it when they have piles of phantom types.

frankmcsherry:
This looks good. I have a few minor doc-requests, which if you could do <3 but I can totally do too:

1. `TimeExchange` description is the same as `Exchange`. 
2. `TimeExchange::new()` says it allocates an `Exchange`.
3. The `exchange_ts` method doesn't describe what it guarantees (which I think is "all records whose time and data evaluate to the same `route` arrive at the same worker" or something like that).

I noticed the build broken due to `RadixSort`. I think I pushed some fixes for that, but perhaps I've screwed that up too. I'll double check!

frankmcsherry:
Ok, I just shoveled a commit at you (at master) that fixes `examples/bfs.rs`. It should be an automatic merge for you, and just fixes an issue of my doing that was causing Travis to complain.

antiguru:
Thanks for fixing the example! I updated the documentation to reflect the differences between `Exchange` and `TimeExchange`.

frankmcsherry:
Looks great, thank you! Once travis comes back, we'll be a-merging. Thanks very much for the contribution. :D

PR #59, opened 2017-04-22
Don't implement Scope for Root

### LINKED ISSUE #53: `Root` probably shouldn't implement `Scope`.
The root of timely dataflow computations currently implements the `Scope` trait, which gives it access to several convenience methods and simplifies the definition of child scopes (which can always point to a parent "scope" to get fresh identifiers and channels and such). At the same time, the root "scope" doesn't do several things that its contained children do, like track communication between operators it hosts.

At the same time, it would help to have more clarity with the root scope for current issues like attaching resources to contained scopes (e.g. for binding libraries lifetimes to those of running dataflow graphs).

The `Root` struct should either behave as a full and complete scope, or it should not implement the trait.  More likely, it should implement a restricted trait related to its role as a dataflow resource allocator.

gereeter:
Note that this is for the `root_rethink` branch.

Fixes #53.

### DIFF
--- a/examples/pingpong.rs
+++ b/examples/pingpong.rs
@@ -1,6 +1,5 @@
 extern crate timely;
 
-use timely::dataflow::Scope;
 use timely::dataflow::operators::*;
 
 fn main() {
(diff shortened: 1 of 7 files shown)


frankmcsherry:
OOOOOOOooo. Thank you! Let me take a peek at this.

frankmcsherry:
This looks good to me! 

There are a few use cases to try out (this was originally in support of writing a dataflow wrapper which could own a shared library, and ensure it is unloaded only after the dataflow the wrapper hosts has completed). I think the changes herein make this a lot easier to sort out, but there are always surprises with Rust. :)

Thanks very much for all the help; do let me know if ever I can be useful back.

PR #61, opened 2017-04-27
Replace PartialOrd with PartialOrder

frankmcsherry:
This PR removes `PartialOrd` as the partial order basis for timestamps in timely dataflow. The trait is problematic in that it has an assumed semantic connection to `Ord`, in that they are assumed by the standard library to have the same implementation when they both exist, preventing a non-total order from implementing `Ord` and doing things like sort-dedup.

Most of the changes here are straightforward, with methods `less_than` and `less_equal` replacing `lt` and `le`. There is a bit of annoying auditing to do, as most timestamp types still derive `Ord` (otherwise they could never be sorted), meaning they have access to `lt` and `le`, and code which uses it as the partial order comparison will not be correct. The `Timestamp` trait does not require `Ord` though, so writing methods generically with `T: Timestamp` should reveal uses of `lt` and `le`.

### DIFF
--- a/src/dataflow/operators/aggregation/state_machine.rs
+++ b/src/dataflow/operators/aggregation/state_machine.rs
@@ -3,6 +3,7 @@ use std::hash::Hash;
 use std::collections::HashMap;
 
 use ::{Data, ExchangeData};
+use order::PartialOrder;
 use dataflow::{Stream, Scope};
 use dataflow::operators::unary::Unary;
 use dataflow::channels::pact::Exchange;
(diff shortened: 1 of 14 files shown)


frankmcsherry:
One thought, that I think I'll try out later today (before merging) is should the PartialOrder methods be part of the Timestamp trait, so that users do not need to include more traits whose only role is for the benefit of Timestamp. This could also allow more "time-idiomatic" naming, e.g. "before" or "prior_to" or "precedes".

frankmcsherry:
Upon further reflection, other things including `PathSummary` need to be partially ordered, not just `Timestamp`s. Moving the functionality into timestamp probably doesn't work for that reason. However, the partial orderings are all time related, so perhaps improving the method names could still help. For now, breaking changes ho!

PR #65, opened 2017-05-05
Deduplicate code in Probe::probe

antiguru:
This PR de-duplicates code in `Probe::probe` by calling `Probe::probe_with` on a newly created handle.


frankmcsherry:
Awesome. I have no clue why I left that there, but this seems totally right. Let's see what the tests say!

PR #66, opened 2017-07-08
Updated intro example code in README to use 'dataflow' instead of 'sc…

drpacman:
…oped' to prevent a compile error when playing along at home


frankmcsherry:
Oh good, thank you. Sorry for that; keeping all the things that aren't rebuilt on check-in in line is hard, apparently. :)

PR #67, opened 2017-08-03
Make Timestamp depend on Hash

aepsil0n:
This is mainly an ergonomics benefit. Since the timestamp in a scope is
often required to be `Hash`, generic code needs to specify that
explicitly. With this change it is implied by the `Timestamp` bound.

Since all existing implementations are `Hash` and the general intention
of a timestamp seems to be something that behaves mostly like an
integer, requiring `Hash` is probably not too much to ask. (Still it's
technically a breaking change.)

### DIFF
--- a/src/dataflow/operators/aggregation/aggregate.rs
+++ b/src/dataflow/operators/aggregation/aggregate.rs
@@ -66,7 +66,7 @@ pub trait Aggregate<S: Scope, K: ExchangeData+Hash, V: ExchangeData> {
 		F: Fn(&K, V, &mut D)+'static, 
 		G: Fn(K, D)->R+'static,
 		H: Fn(&K)->u64+'static,
-	>(&self, fold: F, emit: G, hash: H) -> Stream<S, R> where S::Timestamp : Hash+Eq;
+	>(&self, fold: F, emit: G, hash: H) -> Stream<S, R> where S::Timestamp: Eq;
 } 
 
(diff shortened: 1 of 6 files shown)


aepsil0n:
To elaborate my motivation: I found myself decomposing my dataflow graph into smaller functions operating on generic scopes. And they all ended up with function signatures like this:

```rust
fn iterate<G: Scope>(state: &Stream<G, Grid>) -> Stream<G, Grid>
    where G::Timestamp: Hash
{ /* ... */ }
```

This would be much nicer to write, if all timestamps were `Hash`. As it turns out they are.

With this change the where clause is gone:

```rust
fn iterate<G: Scope>(state: &Stream<G, Grid>) -> Stream<G, Grid>{
    /* ... */
}
```

Is there a good reason to allow unhashable timestamps?

frankmcsherry:
This makes sense. I'm having a hard time thinking of the types that don't implement `Hash` that might be excluded here, but I think the intent of not having `Hash` was just about cutting out traits that the system itself doesn't actually need. In a similar sense it would be great to get rid of `Debug`, but it would also be a huge ergonomic pain in the butt.

There are several existing timestamp types I know of that I suspect do not implement `Hash`, they just aren't part of the timely repository. :)  I don't know any that couldn't `#derive` it, but let me poke the involved people and see if this would be a mess for them.

frankmcsherry:
I got limited feedback, only of the form "sounds like a good idea, actually". So, here we go! :)

PR #68, opened 2017-08-05
input_from introduced

frankmcsherry:
This PR changes `input::Handle` a bit, in particular allowing one to create it independently and then attach it to an input, in much the same way that `probe::Handle` and `probe_with()` work. The intent here is that code can be more idiomatic, with inputs defined and then used, rather than constructed in nested scopes and then returned.

For example, one can rewrite the fragment of `hello.rs` that currently looks like so:

```rust
let (mut input, probe) = worker.dataflow(move |scope| {
    let (input, stream) = scope.new_input();
    let probe = stream.exchange(|x| *x)
                      .inspect(move |x| println!("worker {}:\thello {}", index, x))
                      .probe();
    (input, probe)
});
```

as

```rust

let probe = worker.dataflow(|scope|
    scope.input_from(&mut input)
         .exchange(|x| *x)
         .inspect(move |x| println!("worker {}:\thello {}", index, x))
         .probe()
);
```

or more explicitly as

```rust
let mut probe = ProbeHandle::new();

worker.dataflow(|scope| {
    scope.input_from(&mut input)
         .exchange(|x| *x)
         .inspect(move |x| println!("worker {}:\thello {}", index, x))
         .probe_with(&mut probe);
});
```

These changes, as with the `probe_with` change, are intended to help clarify which things have a dependence on the dataflow scope and which do not. The input handle depends on the data type and timestamp type, but not on the scope. We can even go crazy and use it in multiple scopes, if that is appealing

```rust
let mut probe = ProbeHandle::new();

worker.dataflow(|scope| {
    scope.input_from(&mut input)
         .exchange(|x| *x)
         .inspect(move |x| println!("worker {}:\thello1 {}", index, x))
         .probe_with(&mut probe);
});

worker.dataflow(|scope| {
    scope.input_from(&mut input)
         .exchange(|x| *x)
         .inspect(move |x| println!("worker {}:\thello2 {}", index, x))
         .probe_with(&mut probe);
});
```

The `input::Handle` type is defined so that you can call `input_from` on an active input handle, even one into which data have been sent and whose timestamps have been advanced. Calling `input_from` only provides the records introduced after that call. It is not clear that this functionality is desirable, but you shouldn't be able to break things.

This PR includes a few other changes, and is technically breaking in that it changes the re-export of `operators::handles::InputHandle` as `operators::InputHandle`. This seemed fine internally, though others relying on it may be disappointed. Even if we want to re-export it, we should find a module that is more specific.

This PR also changes `ExchangeExtension` to just be `Exchange`. This had previously clashed with the `Exchange` pact, but .. anyone who wants to use both the `exchange` method and the exchange pact can rebind the pact using `as`.

### DIFF
--- a/examples/hello.rs
+++ b/examples/hello.rs
@@ -1,18 +1,22 @@
 extern crate timely;
 
-use timely::dataflow::operators::*;
+use timely::dataflow::{InputHandle, ProbeHandle};
+use timely::dataflow::operators::{Input, Exchange, Inspect, Probe};
 
 fn main() {
     // initializes and runs a timely dataflow.
     timely::execute_from_args(std::env::args(), |worker| {
+
+        let index = worker.index();
+        let mut input = InputHandle::new();
+        let mut probe = ProbeHandle::new();
+
         // create a new input, exchange data, and inspect its output
-        let (mut input, probe) = worker.dataflow(move |scope| {
-            let index = scope.index();
-            let (input, stream) = scope.new_input();
-            let probe = stream.exchange(|x| *x)
-                              .inspect(move |x| println!("worker {}:\thello {}", index, x))
-                              .probe();
-            (input, probe)
+        worker.dataflow(|scope| {
+            scope.input_from(&mut input)
+                 .exchange(|x| *x)
+                 .inspect(move |x| println!("worker {}:\thello {}", index, x))
+                 .probe_with(&mut probe);
         });
 
         // introduce data and watch!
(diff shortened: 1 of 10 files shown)


utaal:
Neat! Is this (partly) intended for "logging as dataflow"? Regardless, I think this could be useful to connect things: will push a logging PR early next week.

frankmcsherry:
No relation to logging. It was just about making examples a bit easier to read and write. Rather than weird convoluted define then return inputs stuff.

PR #70, opened 2017-08-06
make OutputHandle visible again

kkourt:
Without it, differential-dataflow does not compile.


frankmcsherry:
Oops. Ok, today: figure out what all these handles are and where they should live. Thanks!

frankmcsherry:
PR #71 tries to make this a bit more tasteful. It shouldn't break much (details there; dd re-breaks, but because it relied overmuch of some specific paths in timely), but tries to organize types and traits a bit more hierarchically.

PR #71, opened 2017-08-06
collect generic operators in module

frankmcsherry:
This PR moves various generic operators from the `timely::dataflow::operators` module to one called `generic` nested underneath. All previously visible types and traits exported in `operators` should be re-exported, but named modules are not spoofed (differential dataflow was overly specific about the location of `timely::dataflow::operators::operator::source`, and breaks; there is a branch correcting this that will be merged asap should this land).

Mostly, this PR is an exercise in trying to thin down the `operators` module so that one can more safely do blanket imports. My understanding is that blanket imports are naughty, but everyone wants to type

```rust
use timely::dataflow::operators::*;
```

rather than carefully enumerate all of the extension traits that they actually use. I think ideally we would have several logically coherent modules under `operators`, like `aggregation` is today. Should you want to use the generic operators to write your own, you can get everything you need with just:

```rust
use timely::dataflow::operators::generic::*;
```

Part of the problem was that blanket imports with many names in them seemed to start to lead to silent failures. For example, the `Input` trait was being clobbered by a similarly named type, and silently failing (or, the failures were "cannot find method `new_input()`" rather than "`Input` already defined"). the `InputHandle` type clashed with something else that I wrote, even though I didn't need it imported. 

I think the compartmentalization should make it easier to use classes of methods without worrying about name clashes as much, but I don't have a lot of direct experience organizing extension traits in Rust, so this might be a learning experience (e.g. "everything broke; you suck!").

### DIFF
--- a/src/dataflow/operators/to_stream.rs
+++ b/src/dataflow/operators/to_stream.rs
@@ -4,7 +4,7 @@ use progress::Timestamp;
 
 use Data;
 use dataflow::channels::Content;
-use dataflow::operators::operator::source;
+use dataflow::operators::generic::operator::source;
 use dataflow::{Stream, Scope};
 
 /// Converts to a timely `Stream`.
--- a/examples/barrier.rs
+++ b/examples/barrier.rs
@@ -3,7 +3,8 @@ extern crate timely;
 use timely::dataflow::channels::pact::Pipeline;
 use timely::progress::timestamp::RootTimestamp;
 
-use timely::dataflow::operators::*;
+use timely::dataflow::operators::{LoopVariable, ConnectLoop};
+use timely::dataflow::operators::generic::unary::Unary;
 
 fn main() {
 
--- a/examples/bfs.rs
+++ b/examples/bfs.rs
@@ -9,8 +9,10 @@ use rand::{Rng, SeedableRng, StdRng};
 use timely_sort::{RadixSorter, RadixSorterBase};
 use timely_sort::LSBRadixSorter as Sorter;
 
-use timely::dataflow::operators::*;
+use timely::dataflow::operators::{ToStream, Concat, LoopVariable, ConnectLoop};
+use timely::dataflow::operators::generic::binary::Binary;
 use timely::dataflow::channels::pact::Exchange;
+
 fn main() {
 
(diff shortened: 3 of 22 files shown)


frankmcsherry:
Ping: @utaal, many of whose methods and such are moved around by this.

utaal:
👍 from my side, I wanted to make sense of the generic operators as well. Guess we should also rewrite some of the existing operators::* stuff in terms of the "new" api, at some point.

src/dataflow/operators/to_stream.rs
```diff
@@ -4,7 +4,7 @@ use progress::Timestamp;
 
 use Data;
 use dataflow::channels::Content;
-use dataflow::operators::operator::source;
+use dataflow::operators::generic::operator::source;
```
utaal: Maybe `operators::generic::operator` -> `operators::generic::source`?

frankmcsherry: It could be; I don't think it is re-exported in `generic` yet, but we could do that (to hide where it lives, perhaps)?

frankmcsherry:
Oh I'd totally be in favor of unifying the generic stuff too, but I thought that was likely to break lots of things. The ease of writing the `source` operator has me convinced that the new stuff is the right way to go; I just have to get my brain around it well enough to port the existing `Unary` and `Binary` traits over.

utaal:
Or, I guess, port the operators to the "new" API directly? I.e. use https://github.com/frankmcsherry/timely-dataflow/blob/master/src/dataflow/operators/operator.rs#L68 for something like https://github.com/frankmcsherry/timely-dataflow/blob/master/src/dataflow/operators/aggregation/aggregate.rs ? (This was my intent when writing `unary_frontiers` and family)

src/dataflow/operators/to_stream.rs
frankmcsherry: Ah, it is re-exported in `generic`. I used the longest form possible in all of the examples, under the (possibly misguided) theory that (i) they can easily be corrected if changed, (ii) they test that others like them might break, and (iii) they show others where stuff actually exists.

PR #72, opened 2017-08-06
book contents tested

frankmcsherry:
This PR sets up testing of book contents as part of `cargo test`. It should error if examples no longer build. The examples shouldn't *run*, because .. well I've used the `no_run` decorator and I'm hoping that saves some time and some numbers 0 through 9 in the output. But I've missed at least one so far.

Anyhow, it will be good to have this as part of the test workflow, so that there are always-current documentation about how to build up timely dataflow computations.


frankmcsherry:
Hey, it seems to have worked!

PR #79, opened 2017-08-14
fix formatting in chapter_1_1

shamrin:


frankmcsherry:
Looks good, thanks! The problem, if you are wondering why such a mess, is that the actual output isn't tab-aligned and when dropped into editor-of-choice opinions get expressed. Thank you, though!

PR #80, opened 2017-08-14
propose questions mdbook could probably answer

shamrin:
This is a reply to PR #69:

> Importantly, I would be delighted to accept pull requests on the SUMMARY.md file outlining the structure of the book, as a way of requesting topics. I will be filling things in as they come to me, or as they seem important, but this is a great way to communicate that you'd like a particular area explained.

This PR shouldn't be merged. It's simply a way to ask questions about timely dataflow.

### DIFF
--- a/mdbook/src/SUMMARY.md
+++ b/mdbook/src/SUMMARY.md
@@ -2,6 +2,12 @@
 
 [Timely Dataflow](./introduction.md)
 
+- [Motivation](./chapter_0.md)
+    - Simple but useful example.
+    - What problems does timely dataflow solve?
+    - For what tasks should you use it for?
+    - When shouldn't you use it?
+
 - [Core Concepts](./chapter_1.md)
     - [Dataflow](./chapter_1_1.md)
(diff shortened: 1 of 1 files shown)


frankmcsherry:
These are good points! I forget that not everyone is a high-throughput data junkie, and may have other goals in programming, life. :)

frankmcsherry:
I've pushed some more content of this form. Ideally this would cause this PR to break (because I've tweaked `SUMMARY.md`, but let's see. I suppose the protocol is "close the issue if you find the material fills the intended role". If it nearly does but misses some important features (e.g. I think maybe I should explain the "problems solved" as "managing scheduling, distribution, and serialization" rather than "computational finance" or "biology").

shamrin:
:1st_place_medal: 

Wow, awesome! I haven't finished reading it yet, but it looks like you've answered my questions perfectly. I'll do more GitHub pokes if I find anything.

shamrin:
I've created issue #86 about the things I've found difficult and the things I've found helpful.

PR #82, opened 2017-08-15
Batch progress logic

frankmcsherry:
This PR introduces a large number of breaking changes to the `CountMap` type and the `MutableAntichain` type. The changes are in support of reworking progress-based reasoning to operate more effectively in batch, rather than single updates at a time. The main issue with the previous implementations is that they maintained invariants even for single-element updates, which is expensive, which were being used inside loops over multiple updates.

This removes the single-element update interfaces. These interfaces can be recovered by presenting a single update as an iterator (e.g. `Some(update)`) but are meant to discourage that mode, and to explicitly break things that were using these types that way (sorry!).

There are several other breaking changes, in support of more clarity. The main one is that `CountMap` is now `ChangeBatch`, removing the association with a map (random access to elements) and suggesting that it should be used more as a batch.

There are a few potential improvements: the `drain_into` method on `ChangeBatch` used to just drain the contents of one into the other, but is now able to just do a `mem::swap` if the target is empty. Could help? It's not all horrible.

Anyhow, all of the `ChangeBatch` and `MutableAntichain` public methods are now documented, and each have a test.


frankmcsherry:
Ah, I should also add: both `Notificator` and `FrontierNotificator` got similar work overs. They should have non-breaking changes, in which `notify_at` becomes much cheaper; it used to be a linear scan over the contents, and is now just an append to a list that gets deduplicated later on.

Issue #81, opened 2017-08-15
Capture with `k` workers, replay with `j`, `k ≠ j`


utaal:
`.capture()`-d streams may be sent over sockets to a separate computation.

It's currently unclear how to replay those streams with a different number of workers compared to the source computation.

For example, replaying a 1-worker stream in a j-worker stream, might look like this - but will not work.
```
                let one_stream: Stream<_, Thing> = if index == 0 {
                    one_stream.lock().unwrap().take().unwrap().replay_into(scope)
                } else {
                    vec![].to_stream(scope)
                };
```

The broken progress behaviour seems due to `.to_stream()` expecting to exist on every worker, https://github.com/frankmcsherry/timely-dataflow/blob/eae63eb2063a6e1d3da70c781c3e879a5ccd7d93/src/dataflow/operators/generic/operator.rs#L554

We may just need a source that does not advertise initial capabilities? I'm not sure this checks out with the progress tracking math.

ms705:
Ah, this is good too know about -- I was about to implement a similar paradigm in a situation where _n_ clients talk to _m_ workers (for _n_ != _m_).

I guess it would be useful to have a "multiplex" paradigm, where one dataflow is split into many (for "many" = _m_ in this case). I guess maybe `partition` can be massaged into serving this purpose?

frankmcsherry:
My plan (maybe a bad plan) was to put together a source that could play back arbitrary numbers of sources, from zero to lots, at each worker. The logic isn't too complicated, I think; you still initialize with as many capabilities as there are workers *in the current computation*, and then have each worker maintain the capabilities needed for its collective streams. This might mean immediately dropping the capability if it has no streams, and it might mean a `MutableAntichain` to track the lower bounds if it has multiple streams.

I'm not sure the best way to do this generally, for `replay` as well as other sources of inputs (I'm looking at reading Kafka queues right now, for example), but maybe doing it once we can look at the result and see if we like it?

frankmcsherry:
More generally, if you want to have multiple "proxy workers" managed by one worker, it should be pretty easy: each worker in most *source* operators apply a constructor where they get a capability, and produce a closure that acts on the output handle.

It seems pretty easy to rig a single worker to manage multiple workers *for sources*: in initialization you clone your capability as many times as you have workers to manage, invoke each of their constructors and assemble their closures in a `Vec`. When you are invoked with each handle, you call each of your stashed closures on the output handle.

[...]

PR #84, opened 2017-08-16
Capture replay

frankmcsherry:
This PR fixes at least one bug, and substantially generalizes the abilities of capture and replay. It breaks some external interfaces, but following in proud Rust tradition this is fine because it was buggy in the first place.

The bug fixed is that capture and replay should not have been correctly working with multiple workers. The examples and tests were not seeing anything, in part because the replayed stream was just `inspect`ed rather than followed by an operator which required correct frontier information. The symptoms should have been premature frontier advancement, which totally happens to lots of systems. It should be fixed now.

At the same time, `replay_into` is now implemented for *iterators* of `EventIterator` implementors, rather than just for the `EventIterator` type. This mean that you replay *collections* of streams, rather than individual streams. This is a substantial increase in flexibility, as it means that you can replay streams captured with some number of workers in a dataflow with a different number of workers (more, fewer, or the same).

The `operators::capture` module is now a directory rather than a monolithic file, with parts broken out into several pieces and possibly better documented. There are not full doctests throughout, which is a good future goal.

There is more documentation about the "replay protocol", meaning what sequence of `Event`s an `EventIterator` should produce, in the `capture::replay` module-level documentation. The rough version is that it is the sequence of progress changes and messages received, *starting from an initial `(Default::default(), 1)` capability*. So, you could imagine a sequence that looks like

    Event::Progress([(0, -1), (5, +1)]),
    Event::Message((5, ["hi there"])),
    Event::Progress([(5, -1)]),

Notice that there is no initial `(0, +1)`; this is assumed, and it is important that it is assumed (the replayers describe their own initial capabilities, and create the initial capabilities from the replayed streams based on their number: there is one for each stream, so they duplicate up the capability).

Otherwise, the sequence of events proceeds as you might imagine, decrementing capabilities and sending data only at times for which the capability exists. The stream should end by releasing all of its capabilities.

Each replayer just replays the instructions as provided, merging all streams that it manages without thinking very hard (any interleaving should be valid as long as each stream remains in order). This makes it relatively easy to replay arbitrary subsets of streams, without requiring a sophisticated replayer. A smarter one could pull from the iterators it thinks would make the most progress, that sort of thing, but this can be future work.


frankmcsherry:
This PR also changes the examples `capture_send.rs` and `capture_recv.rs` to support multiple workers (it should break on multiple computers, because of how it tries to connect streams with TCP to localhost). The `capture_recv` example is a bit more complicated, but demonstrates how a different number of workers can be used to replay streams; it takes the number of source workers as an argument, sets up that many TCP listeners, and plays back that many streams, using the number of workers supplied by the `-w` parameter.

frankmcsherry:
I think I'm supposed to type "fixes #81".

tschottdorf:
You'll need "Fixes #81." in its own line in either PR description or commit message.

PR #87, opened 2017-08-18
Misc comments on Progress tracking

tschottdorf:
Feel free to close when you have read them, @frankmcsherry.

### DIFF
--- a/mdbook/src/chapter_4/chapter_4_2.md
+++ b/mdbook/src/chapter_4/chapter_4_2.md
@@ -2,16 +2,17 @@
 
 Progress tracking is a fundamental component of timely dataflow, and it is important to understand how it works to have a complete understanding of what timely dataflow does for you.
 
-Let's start with a statement about what progress tracking means to accomplish. The setting is that there are multiple workers, each of whom move timestamped data through a common dataflow graph. The data may move between workers, and as the data are processed by operators we have relatively few guarantees about the consequences: mostly that the resulting timestamps can only move forward. Nonetheless, the workers engage in a collective protocol that ensures that each worker is able to report, to each location in its hosted dataflow graph, the impossibility of ever observing certain timestamps at that location again.
+Let's start with a statement about what progress tracking means to accomplish. The setting is that there are multiple workers, each of whom move timestamped data through a common dataflow graph. The data may move between workers, and as the data are processed by operators we have relatively few guarantees about the consequences: mostly that the resulting timestamps can only move forward [it's not obvious what this means]. Nonetheless, the workers engage in a collective protocol that ensures that each worker is able to report, to each location in its hosted dataflow graph, the impossibility of ever observing certain timestamps at that location again.
 
 As an overview, timely dataflow's progress tracking is about (i) maintaining a view of *timestamp capabilities* at each location in the dataflow graph, and (ii) communicating the implications of changes in capabilities to other locations in the dataflow graph. Before we get in to these two aspects, we will first need to be able to name parts of our dataflow graph.
 
 ## Dataflow structure
 
-A dataflow graph hosts some number of operators. For progress tracking, these operators are simply identified by their index. Each operator has some number of *input ports*, each of which may independently have or not have certain timestamps in its future. Each operator has some number of *output ports*, each of which may independently express a capability to produce timestamps in the future. Each input port is connected to a single output port, and each output port may be connected to multiple input ports (a message produced at an output port is to be delivered to all attached input ports).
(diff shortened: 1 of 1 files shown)


mdbook/src/chapter_4/chapter_4_2.md
```diff
@@ -2,16 +2,17 @@
...
 
-In timely dataflow progress tracking, we identify output ports by the type `Source` and input ports by the type `Target`, as an output port is a source of timestamped data, and an input port is a target of timestamped data. Each can be described by their operator index and then an operator-local index of the corresponding port. The use of distinct types helps us avoid mistaking input and output ports.
+In timely dataflow progress tracking, we identify output ports by the type `Source` and input ports by the type `Target`, as an output port is a source of timestamped data, and an input port is a target of timestamped data[this nomenclature is confusing, but hopefully less so in the actual code? Would be weird if operators used their "source" to write to]. Each can be described by their operator index and then an operator-local index of the corresponding port. The use of distinct types helps us avoid mistaking input and output ports.
 
+[hmm, I would really think that data comes from Source and goes to Target. Is it really the other way?]
```
frankmcsherry: It is, because we are talking about all of this from the perspective of the manager of the dataflow graph. It doesn't know how the operators' internals work, just that the outputs of operators produce data, and the input consume data. Unmentioned in the text, there are other sources and targets too (e.g. in a nested scope, the "ingress" nodes are sources, and the "egress" nodes are targets. They look like operator outputs and inputs, respectively.

mdbook/src/chapter_4/chapter_4_2.md
```diff
@@ -46,9 +47,10 @@ As a computation proceeds, operators may perform three classes of action:
...
+always correspond to information corresponding to some union of "data batches"?]
 
-Each worker reliably broadcasts the stream of progress change batches to all workers in the system (including itself) along FIFO channels. At any point in time, each worker has seen an arbitrary prefix of the sequence of progress change batches produced by each other worker, and it is expected that as time proceeds each worker eventually sees every prefix of each sequence.
+Each worker reliably broadcasts the stream of progress change batches to all workers in the system (including itself) along FIFO channels. At any point in time, each worker has seen an arbitrary prefix of the sequence of progress change batches produced by each other worker, and it is expected that as time proceeds each worker eventually sees every prefix of each sequence[What ordering guarantees do you have here? This looks like the kind of thing that's annoying to get completely right in practice].
```
frankmcsherry: The ordering guarantees should be just that each point-to-point (worker to worker) progress update channel is FIFO and at most once. If you make it "reliable" then you get progress too. There is no requirement that these messages be correctly interleaved with data or anything like that. In fact, if you do some silly looking changes to the progress protocol, you can drop the FIFO requirement (if you change it to disallow cloning of capabilities, and require consuming capabilities when sending a message; this would mean building in a magical 100000x inflation on received capabilities, or something similar, but it would remove the requirement that progress update batches be FIFO (and later in the text "prefix of progress update batches" could be replaced with "subset of progress update batches" ).

mdbook/src/chapter_4/chapter_4_2.md
```diff
@@ -77,7 +79,7 @@ pub trait PathSummary<T> : PartialOrder {
 
 The types implementing `PathSummary` must be partially ordered, and implement two methods: 
 
-1. The `results_in` method explains what must happen to a timestamp moving along a path (note the possibility of `None`; a timestamp could *not* move along a path, for example a timestamp with a loop counter that is about to exceed the loop limit will be discarded rather than advanced).
+1. The `results_in` method explains what must happen to a timestamp moving along a path (note the possibility of `None`; a timestamp could *not* move along a path, for example a timestamp with a loop counter that is about to exceed the loop limit will be discarded rather than advanced).[OTOH, if the loop isn't about to terminate, then morally `(1,2).results_in.unwrap() == (1,3)`].
```
frankmcsherry: Right, so you could have a summary for which 

```rust
(1,2).results_in() == Some((1, 3));
(1,u64::max_value).results_in() == None;
```

mdbook/src/chapter_4/chapter_4_2.md
```diff
@@ -135,11 +137,11 @@ Second, just before a worker might hypothetically receive a message at `(l1, t1)
 
 At this point, we have two important bits of knowledge:
 
-1. The accumulated values for all pointstamps less than `(l1, t1)` were initially non-positive, by assumption.
+1. The accumulated values for all pointstamps less than `(l1, t1)` were initially non-positive, by assumption.[because everything is settled except the inflight message, which means someone has decremented for it, but the receiver hasn't incremented?]
```
frankmcsherry: No, this "by assumption" is meant to be "by the assumption of the safety property". Perhaps it is too tangled up in the text, but the claim is that "if there are no pointstamps with positive count that could-result-in `(l1,t1)`, then ..."

frankmcsherry:
These are helpful, thank you. I put some comments in where I thought they might be informative (and as exercise in explaining them), but I've got all of them marked as "problematic"

frankmcsherry:
I think I've gotten all of these, so closing in the spirit of the request. There *is* more to write, though, so perhaps there will be more issues too. :)

PR #88, opened 2017-08-19
Split out a SubgraphBuilder type that exists only during the dataflow construction phase.

gereeter:
This splits out a `SubgraphBuilder` type from `Subgraph`, which allows representing things differently and doing initialization routines while building the dataflow graph without worrying about interleaving those things with actual execution.

### DIFF
--- a/src/progress/nested/subgraph.rs
+++ b/src/progress/nested/subgraph.rs
@@ -49,6 +49,38 @@ pub struct Target {
     pub port: usize, 
 }
 
+/// A builder structure for initializing `Subgraph`s.
+///
+/// This collects all the information necessary to get a `Subgraph` up and
+/// running, and is important largely through its `build` method which
+/// actually creates a `Subgraph`.
+pub struct SubgraphBuilder<TOuter: Timestamp, TInner: Timestamp> {
+    /// The name of this subgraph.
+    pub name: String,
+
+    /// A sequence of integers uniquely identifying the subgraph.
+    pub path: Vec<usize>,
+
+    /// The index assigned to the subgraph by its parent.
+    pub index: usize,
+
+    inputs: usize,
+    outputs: usize,
+
+    // handles to the children of the scope. index i corresponds to entry i-1, unless things change.
+    children: Vec<PerOperatorState<Product<TOuter, TInner>>>,
+    child_count: usize,
+
+    edge_stash: Vec<(Source, Target)>,
+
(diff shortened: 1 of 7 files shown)


frankmcsherry:
This sounds pretty exciting. :)  I'll take a look asap!

src/progress/nested/subgraph.rs
```diff
@@ -491,6 +506,103 @@ impl<TOuter: Timestamp, TInner: Timestamp> Operate<TOuter> for Subgraph<TOuter,
...
+        self.child_count - 1
+    }
+
+    /// Adds a new child to the subgraph.
+    pub fn add_child(&mut self, child: Box<Operate<Product<TOuter, TInner>>>, index: usize, identifier: usize) {
+        self.children.push(PerOperatorState::new(child, index, self.path.clone(), identifier))
+    }
+
+    /// Now that initialization is complete, actually build a subgraph.
+    pub fn build<A: Allocate>(mut self, allocator: &mut A) -> Subgraph<TOuter, TInner> {
```
frankmcsherry: One very appealing consequence of this signature is that all of the channels in a subgraph could be allocated at this point, in principle. That is, one can whip up a subgraph builder that describes the structure of the computation underneath it, but is agnostic to the actual numbers of workers / their current location / etc. Only when this method is called do we construct the actual connections even further down.

Maybe a different way to state: you could imagine somewhere a trait 

```rust
pub trait OperateBuilder<T: Timestamp> {
    fn build<A: Allocate>(mut self, allocator: &mut A) -> Box<Operate<T>>;
}
```

and one could in principle get all of a subgraph ready to go with operator builders in place, but with zero hooks into a running timely system (so "just data" at the moment), and then instantiate everything only once asked.

We did a similar thing in Naiad, but I think there was more of a start-up sequence there. I'll try to page back in why it was so important.

frankmcsherry:
Yeah, this all looks pretty great to me. My debt continues to mount, and I am forced to contribute to other open-source projects to pay it down. :)

tschottdorf:
> Yeah, this all looks pretty great to me. My debt continues to mount, and I am forced to contribute to other open-source projects to pay it down. :)

I'm curious, what do you mean?

frankmcsherry:
@gereeter has dropped several other useful PRs on me, but remains elusive, and I have no idea where to order drinks delivered / kudos awarded.

PR #90, opened 2017-08-23
Generic operators

frankmcsherry:
This PR is probably not ready to merge yet (it passes tests on my laptop, but hasn't been exercised with multiworker/multiprocess runs).

This reworks `dataflow/generic/operator.rs`, something that @utaal built a while back to be a more general implementation of generic operators (as compared with `unary.rs` and `binary.rs`, which each share some redundant code).  This goes even further, and provides an `OperatorBuilder` type that should be able to build operators with arbitrary numbers of inputs and outputs, programmatically.

Let's start with an example. The `binary_frontier` method from `operator.rs` takes two streams with frontier information, and runs whatever code you supply. The logic to make one of these used to be pretty hairy (you can check it out in the current `operator.rs`. With the new operator builder it is much simpler, and I think almost approachable.

**Edit**: updated to reflect current version.

```rust
let mut builder = OperatorBuilder::new(name.to_owned(), self.scope());

let mut input1 = builder.new_input(self, pact1);
let mut input2 = builder.new_input(other, pact2);
let (mut output, stream) = builder.new_output();

builder.build(move |capability| {
    let mut logic = constructor(capability);
    move |frontier| {
        let mut input1_handle = new_frontier_input_handle(&mut input1, &frontier[0]);
        let mut input2_handle = new_frontier_input_handle(&mut input2, &frontier[1]);
        let mut output_handle = new_output_handle(&mut output);
        logic(&mut input1_handle, &mut input2_handle, &mut output_handle);        
    }
});

stream
```

The idea here is that `OperatorBuilder` puts together the state that the operator needs to know about, which is mostly just several `Rc<RefCell<ChangeBatch<T>>>` shared lists of progress information. As you call `new_input` and `new_output`, it mints up the data movement you need, and holds on to the progress information it needs. When you finally call `build`, it zips itself up and installs itself in the scope that it initialized with.

There are a few caveats, mostly that you still need to be smart enough to grok the contracts of the various handles that get minted up. They should be simpler now, as the `output_handle` type up above now calls `self.cease()` in its `Drop` implementation, so you don't have to worry about that (was a serious correctness risk). 

It is important to use `internal` for your input handles so that minted capabilities are registered; failing to do this will cause the progress tracker to go sideways. It may make sense to have input handles own a ref-counted `ChangeBuffer` that is just `internal`, so that no one can screw this up (other than us).

I'm going to put it through some more tests next, but I'm hopeful that this should simplify operator implementation for many folks, for example @aepsil0n's possible need for a two-input, two-output operator to switch between iterating and dropping data out of a loop, based on a control input. 

Any comments are great, of course!


frankmcsherry:
Some more food for thought on options:

1. In the example above, there are `Rc<RefCell<ChangeBatch<T>>>` types sprinkled liberally so that the input and output handles don't need to synchronize with the operator. We could do the same thing with `internal`, and there we be relatively few ways to "go wrong". On the down-side, we've put an indirection (via the `Rc`) in the way of each buffer access, including for operators with nothing done.

2. We could alternately have implementations with `&'a mut ChangeBatch<T>` in them instead, which must be explicitly and correctly constructed, but which avoid the indirection. There is some ergonomic pain (all of the `ChangeBatch`es are in one allocation, so lots of `split_mut` to break out the disjoint references), but it avoids accumulating minor costs.

3. We could do both. Neither are very complicated, but involve fusing in some buffering on the output side (the counter is post-buffering).

frankmcsherry:
These new commits re-target the existing generic operators (unary.rs, binary.rs, and operator.rs) to use the common operator builder. Each of the implementations are now relatively concise. Their signatures have not changed, but some details of the types in the signatures have changed (e.g., `Notificator` now has a lifetime, as it borrows its frontier information rather than owns it). They should largely be compatible, unless your code was very explicit with its types.

Several other collateral changes happened. Perhaps the most significant is that `ParallelizationContract`s now have `Pusher` and `Puller` associated types, which are the types they return when allocating channels. There was previously a great deal of boxing going on, in the interest of reducing generic parameters / staying sane, but I think all of this has been removed except at `Tee`, where the heterogenous nature of the `Push` implementors means we need to box things.

frankmcsherry:
This was interesting. The new operator logic starts all operators with capabilities and has the logic then decline the capability if they need. This means that lots of operators have capabilities, something that wasn't much tested before, it seems. There was a bug in the subgraph's `get_internal_summary` method that caused subgraphs not to report their internal capabilities correctly (at all). Should be better now.

frankmcsherry:
As part of this, I've added two tests in `subgraph.rs` that check the validity of operator progress statements. They are guarded by a `#[cfg(debug_assertions)]` and should only be active in debug mode, but they should help raise confidence in tests.

They are statements about the `consumed_buffer`, `internal_buffer`, and `produced_buffer` fields that hold proposed changes by the operator. The invariants they need to maintain relate to each other, and to the capabilities held by the operator before the change (in `self.internal`).

1. For each output, each increment to a held capability must correspond to a consumed message or an existing capability at a less or equal time.
    ```rust
    for index in 0 .. self.internal_buffer.len() {
        for change in self.internal_buffer[index].iter() {
            if change.1 > 0 {
                let consumed = self.consumed_buffer.iter_mut().any(|x| x.iter().any(|y| y.1 > 0 && y.0.less_equal(&change.0)));
                let internal = self.internal[index].less_equal(&change.0);
                if !consumed && !internal {
                    panic!("Progress error; internal {:?}: {:?}", self.name, change);
                }
            }
        }
    }
    ```
2. For each output, each produced message must correspond to a held capability or a consumed message at a less or equal time.
    ```rust
    for index in 0 .. self.produced_buffer.len() {
        for change in self.produced_buffer[index].iter() {
            if change.1 > 0 {
                let consumed = self.consumed_buffer.iter_mut().any(|x| x.iter().any(|y| y.1 > 0 && y.0.less_equal(&change.0)));
                let internal = self.internal[index].less_equal(&change.0);
                if !consumed && !internal {
                    println!("internal: {:?}", self.internal[index].frontier());
                    panic!("Progress error; produced {:?}: {:?}", self.name, change);
                }
            }
        }
    }
    ```

It turns out these are basically the same test, because our ability to mint a capability is equivalent to our ability to send a message. Perhaps that should tell us something about how to simplify the protocol (no, we cannot merge capability counts and produced messages).

Anyhow, these tests are live for every operator call in debug, including subgraphs and such, but are not included when compiled with release mode.

PR #98, opened 2017-10-13
PACT: Allow to make the output ports of Exchange configurable

antiguru:
So, this is an ugly one. It's my poor attempt to tell operators to not blindly throw data downstream but get them to do it in a more ordered way, i.e. provide a map of downstream workers. This might totally screw up progress tracking and rock Timely's foundations, but at least one example does not crash. In research, we call it success.

It reads an environment symbol called WMAP, which is supposed to have many ID:MAP entries separated by a single space. The ID refers to the exchange, that's why there's a nasty print statement in there. The MAP is a u64 where a set bit indicates, yes, send data to the n-th downstream worker. And who needs more than 64 workers, anyways.

Enjoy it with `WMAP="5:11" cargo run --example pingpong -- 5 -w 16` It still counts to 5 although doesn't tell anyone.

I'm mostly interested in your opinion on this, if it is the right way of implementing it without breaking other things and if there's a better way to figure out what ID an exchange has.

Cheers

### DIFF
--- a/src/dataflow/channels/pact.rs
+++ b/src/dataflow/channels/pact.rs
@@ -63,8 +63,28 @@ impl<T: Eq+Data+Abomonation, D: Data+Abomonation, F: Fn(&D)->u64+'static> Parall
     type Pusher = Box<Push<(T, Content<D>)>>;
     type Puller = Puller<T, D, Box<Pull<Message<T, D>>>>;
     fn connect<A: Allocate>(self, allocator: &mut A, identifier: usize) -> (Self::Pusher, Self::Puller) {
+        println!("Exchange::connect identifier: {:?} allocate.index: {:?} allocate.peers: {:?}", identifier, allocator.index(), allocator.peers());
         let (senders, receiver) = allocator.allocate::<Message<T, D>>();
-        let senders = senders.into_iter().enumerate().map(|(i,x)| Pusher::new(x, allocator.index(), i, identifier)).collect::<Vec<_>>();
+        let senders = if let Ok(wmap) = ::std::env::var("WMAP") {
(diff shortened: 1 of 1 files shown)


frankmcsherry:
To see if I understand, the intent here is to subset out the workers, and only distribute the records among that subset, is that correct?

We've done similar things in the past, though without environment variables and such. Let me explain and you can tell me if this is similar to what you want or not.

If we rely on the property that records are distributed in exchange channels by taking the function applied to each record and then the modulo number of workers, we can explicitly route records by hijacking the distribution function. Imagine you wanted to route records to a set of workers described by a `targets: Vec<usize>`, we could write

```rust
let exchange = Exchange::new(move |x| targets[hash(x) % targets.len()])
```

and use this pact to route records as if there were only the endpoints in `targets`, exploiting the property that for `i < workers`, the value `i` routes data to worker `i`.

We can also do other tricks, like a "process-local exchange", in which we exchange data using a function that takes `hash(x) % local_workers` and adds the offset that lands the data on the local process (usually `local_workers * (worker.index() / local_workers)`). This effects process-local data exchange, without inventing any new infrastructure. There could be good reason to do specialized process-local data exchange (e.g. neither communicating nor blocking on progress information outside  the process), but this is a quick fix in a similar spirit, I think.

How does this look to you? Would this type of approach solve the use cases you have in mind?

antiguru:
Yes, our use case is to be able to set different operator parallelism. I guess what you suggest would work quite as well. I'll test it and see if it makes any difference (Looking at the code the main obstacle seems modulus versus a mask.)

antiguru:
So, I'll close this because I'll need a more thorough mechanism. Will create a new PR once I have something!

PR #103, opened 2017-11-03
fix typos in mdbook

vasia:
I really enjoyed reading ^^


frankmcsherry:
Thank you, both for your nice words and the list of my not nice words.

PR #104, opened 2017-11-03
Fix typo in mdbook

Moredread:


frankmcsherry:
Looks great; thank you very much for the attention to detail!

PR #105, opened 2017-11-04
Order mdbook SUMMARY.md by chapter number

adelbertc:
I generated a PDF to read by ordering the files by filename. When I got to chapter 4 I was wondering why I saw "Internals" before "Advanced Timely Dataflow" when the latter appeared before the former in SUMMARY.md.. then I realized :-)


frankmcsherry:
Ah, so the story here is actually the other way around. I think the advanced topics are where they are supposed to be, but they were added after the internals were. So, probably the right fix is to re-number the advanced topic stuff to be section 4 and internals to be 5. Or perhaps to remove the numbering (in case this happens again)? I can do either of these, but which of these fixes makes the most sense from your point of view; was the numbering helpful for the pdf point of view / would it help to have a .pdf committed?

adelbertc:
I can see Advanced Timely Dataflow coming before Internal, makes sense to cover all the user guide stuff before diving into internals - I'll amend this PR with that change.

PDF might be tricky since you'd want to keep it in sync with the markdown which involves some ceremony with Travis, but if there's interest I can try to wire some stuff up. Maybe something like have Travis build it and push to a GitHub pages branch on merge to master.

frankmcsherry:
Oops; I didn't realize that you had pushed more commits at this. Looks good to me!

PR #116, opened 2018-01-06
Work in progress: Pointstamp tracking

frankmcsherry:
This PR is a work in progress for refactored progress tracking. Currently the progress tracking state is scattered around a few elements of `subgraph.rs`, where it could be better isolated so that the components can be understood, tested, and perhaps replaced with alternate implementations.

The current intent is that the `PointstampCounter` type, which currently maintains buffers of raw and pushed pointstamp updates, is combined with the path summaries currently held by operators. These two together allow the `PointstampCounter` to expose methods

```rust
    pub fn update_target(&mut self, target: Target, time: T, value: i64);
    pub fn update_source(&mut self, source: Source, time: T, value: i64);
    pub fn push_pointstamps(&mut self);
    pub fn pushed_mut(&mut self, node: usize) -> &mut [ChangeBatch<T>];
```

The first two methods allow a user to report observed changes at targets and sources (input and output ports, respectively), and the second two methods allow a user to propagate and observe the implications of these changes elsewhere in the graph. These methods seem like a reasonable basis for a trait, perhaps with some more abstraction so that we can efficiently report changes rather than requiring the user to effectively poll nodes and ports for changes.

This `PointstampCounter` is constructed by an associated `Builder`, which has methods

```rust
    pub fn add_node(&mut self, index: usize, summary: Vec<Vec<Antichain<T::Summary>>>);
    pub fn add_edge(&mut self, source: Source, target: Target);
    pub fn build(&mut self) -> PointstampCounter<T>;
```

These methods seem roughly in line with what we need in `subgraph.rs`, though the PR does not yet swap them in to check this (more commits will be pushed before this happens).


frankmcsherry:
I've pushed a bit of documentation and tests. I don't have a great sense for how to test this thoroughly (there are an unbounded number of weird graph shapes that I couldn't possibly think of). I'm using the following in the documentation as a running example, though:

```rust
use timely::progress::frontier::Antichain;
use timely::progress::nested::subgraph::{Source, Target};
use timely::progress::nested::reachability::{Builder, Tracker};

// allocate a new empty topology builder.
let mut builder = Builder::<usize>::new();
 
// Each node with one input connected to one output.
builder.add_node(0, 1, 1, vec![vec![Antichain::from_elem(0)]]);
builder.add_node(1, 1, 1, vec![vec![Antichain::from_elem(0)]]);
builder.add_node(2, 1, 1, vec![vec![Antichain::from_elem(1)]]);

// Connect nodes in sequence, looping around to the first from the last.
builder.add_edge(Source { index: 0, port: 0}, Target { index: 1, port: 0} );
builder.add_edge(Source { index: 1, port: 0}, Target { index: 2, port: 0} );
builder.add_edge(Source { index: 2, port: 0}, Target { index: 0, port: 0} );

// Construct a reachability tracker.
let mut tracker = Tracker::allocate_from(builder.summarize());

// Introduce a pointstamp at the output of the first node.
tracker.update_source(Source { index: 0, port: 0}, 17, 1);

// Propagate changes; until this call updates are simply buffered.
tracker.propagate();

// Propagated changes should have a single element, incremented for node zero.
assert_eq!(tracker.pushed_mut(0)[0].drain().collect::<Vec<_>>(), vec![(18, 1)]);
assert_eq!(tracker.pushed_mut(1)[0].drain().collect::<Vec<_>>(), vec![(17, 1)]);
assert_eq!(tracker.pushed_mut(2)[0].drain().collect::<Vec<_>>(), vec![(17, 1)]);
```

frankmcsherry:
The commit just above this runs both progress tracking implementations and asserts that they match. All tests pass, though they didn't before a few bug fixes. I'm going to try this out with a few more complex dataflows (differential), and once more confidence is established the old implementation comes out.

frankmcsherry:
The code currently provides a `Tracker` which supports the same operations as before (updates at sources and targets, and access to a `&mut [ChangeBatch<T>]` to read out the progress implications). Unlike previous versions, this one internally maintains `MutableAntichain<T>` instances for each source and target, so that it performs the "set of updates" -> "frontier" operation that is mandatory for correct execution.

This does leave some redundant work going on in the computation at the moment (the conversion to a frontier is done twice, outside and inside the progress tracker). In the near future this should be fixed, but the path to removing the redundant operations seems to involve a few other structural changes to (the outside instances track information that is used to "ensure" correctness for a slightly broken api, which should just be fixed at which point they can be removed).

PR #118, opened 2018-01-07
Connection builder connection setup

antiguru:
This pull request implements an extended API for the builders. Each `new_input` and `new_output` function has a corresponding `_connection` variant, which allows to specify a vector of `Antichain` of connection information.
The connection information can be used to explain to Timely how timestamps are passed from inputs to outputs. A connection is created using a `Antichain::from_elem(Default::default())` value, no connection is indicated by an empty antichain (`Antichain::new()`).


frankmcsherry:
Looks good to me! Thanks very much. :D

PR #119, opened 2018-01-09
Notificator frontier

antiguru:
`Notificator::frontier` returns `&MutableAntichain` instead of `&[]`.


frankmcsherry:
I'm pondering this a bit, and I wonder if it might make sense to have a `&[T]` wrapper called something like `AntichainRef` (name optional) that is just a wrapper type providing the guarantee that a slice of times are an antichain, and providing methods like `less_equal`, `less_than`, etc. Would that serve your purposes just as well?

antiguru:
Yes, that would be helpful also in other places where one has a `[T]` and one is sure that it represents the elements of an antichain. I feel generalizing it could definitely be helpful.

PR #126, opened 2018-01-26
Operator identifier is exposed to operator constructor

utaal:


frankmcsherry:
Looks good to me. Seems pretty easy to fix inevitable breaks. :)

PR #136, opened 2018-03-01
Cleanup

emanuelbutacu:
This PR is about cleaning the code. It includes the following:

- normalize the final newline
- remove trailing whitespace
- use spaces as indentation
- remove empty files

PR #137, opened 2018-03-01
Clippy

emanuelbutacu:
This PR is about implementing most of the changes that clippy suggested.


frankmcsherry:
So, I have a few issues with some of the classes of clippy suggestions. I think there are some good ones to take here (I try to keep up with the "don't take double references", and I think the "don't need `field: field` initializers" landed since I last ran it). At the same time, there are a bunch that I currently disagree with (e.g. "use `&iter`" instead of `iter.iter()`", convert various `|| Vec::new()` to `Vec::new`, other places where cuteness trumps explanatory idioms).

I'm up for either taking a modified PR (after chatting through what's good / bad), or doing a clean-up pass myself and giving out an ack. Any thoughts?

emanuelbutacu:
In that case, I'll update the PR. From what I understand, we want to remove the following changes:

- all the iterator conversions (mutable versions too), eg. `iter.iter()` to `&iter`
- all closure conversions, eg. `|| Vec::new()` to `Vec::new`

Should I include something else?

frankmcsherry:
I took a swing through, and I would vote to *include*: 

1. don't take double references,
2. simplify `field: field` to `field`,
3. use type names in doc text (e.g. "`T`" vs "T"),
4. sort derived impls alphabetically,

I would exclude:

1.  Use cutesy IntoIter impls for &iter, &mut iter,
2. Short cut function calls (e.g. `|| Vec::new()` to `Vec::new`) and unary enum constructors,
3. Replace `.clone()` calls with derefs for things implementing `Copy`,
4. Remove lifetimes when lifetime elision would work,
5. Some weird rule about "accept slice refs and then call `to_owned`, rather than `Vec::clone`",
6. Add `Default` impls whenever possible,
7. whatever proposes this change:
```rust
 -            for node in 0 .. target_target.len() {
 +            for (node, _) in target_target.iter().enumerate() {
```
8.. whatever proposes to always replace `0 .. x.len` iteration with `x.iter()` (there are plenty of times where the former is done to better explain what we are iterating over),

So, clearly I have opinions. I think the opinions are held more strongly by being told by some clippy team members that my opinions are wrong, and that I will eventually realize that. :)

But, I'd go for whatever is easiest for you, and ideally you don't take offense if in the future one or two of the revisions drifts back in in what is probably an attempt to keep relatively tricky code as clear as possible.

emanuelbutacu:
I updated the PR with the changes you suggested. Please let me know if I missed anything from your suggestions.

frankmcsherry:
It looks like there are now some conflicts, probably with the accept of your #136 clean-up PR. Would you mind taking a look at them? I checked through the diffs against the prior code, and these changes seem good otherwise!

emanuelbutacu:
I am looking into them now.

frankmcsherry:
Looks good to me. Let's let Travis confirm, and then should be good to merge.

frankmcsherry:
Thanks very much!

emanuelbutacu:
Thanks for bearing with me.

frankmcsherry:
No worries; thank you for helping to keep everything tidy!

PR #139, opened 2018-03-13
logging: Start with a non-empty frontier in EventPusherTee

### LINKED ISSUE #138: Logging delivers incorrect progress information for logging stream
The logging event stream uses nanosecond timestamps as capabilities. As the default capability for the type `u64` is 0, the first progress information send to a subscriber is to drop the default capability. However, this causes the subscriber to believe that it is done and it might quit prematurely.

The solution to the problem is to only drop a default capability when a new one exists, i.e. when logging data has been received.

antiguru:
Starting with an empty frontier causes the default capability to be
dropped twice on the log consumer.

Fixes #138.

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -4,6 +4,7 @@ use std::rc::Rc;
 use std::cell::RefCell;
 use std::sync::Arc;
 
+use ::progress::change_batch::ChangeBatch;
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Looks good to me (and appears to fix actual observable breakage)! Thank you very much!

PR #140, opened 2018-03-14
Implement AntichainRef type

antiguru:
In this PR, I'm implementing the type `AntichainRef` wrapping a slice of timestamps that are elements of an antichain, i.e. mutually incomparable elements. It is a convenience wrapper to provide functions such as `less_than`, `less_equal` to aid programmers using frontier semantics in operators.


frankmcsherry:
Yeah, this looks good to me. Hard to see any downside, so we are doing it!

PR #146, opened 2018-04-11
Re-export timely_communication::Allocate as timely::Allocate

antiguru:
Re-export timely communication's `Allocate` trait as `timely::Allocate` because it is part of the `ParallelizationContract`'s public API.


frankmcsherry:
This seems totally sane. Not sure why we shouldn't do this. Merging until someone explains that it was bad. :)

PR #150, opened 2018-04-16
Remove getopts

qmx:
this is the initial work addressing https://github.com/rust-lang/rust/issues/49925

I've removed the dependency from where it isn't used directy, feature-gated the uses, enabled it by default

Feedback is more than welcome :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,6 @@ time="0.1.34"
 
 [dev-dependencies]
 timely_sort="0.1.6"
-getopts="0.2.14"
 rand="0.4"
 #skeptic = "0.12"
 
--- a/communication/Cargo.toml
+++ b/communication/Cargo.toml
@@ -13,7 +13,7 @@ keywords = ["timely", "dataflow"]
 license = "MIT"
 
 [features]
-default=[] # ["logging"]
+default=["arg_parse"] # ["logging"]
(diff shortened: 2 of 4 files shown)


frankmcsherry:
This seems sane. Do you have any thoughts on a feature used negatively, like

```rust
#[cfg(not(feature = "hide_get_opts"))]
```

I guess all the negation is more annoying to grok, but I'm trying to think about how to minimize the cognitive load of what "default" is (and e.g. what to do if there is another crate we want to optionally suppress; `arg_parse` is probably less optimal than e.g. `timely_core`, maybe?).

This is kind of bikeshedding; the config seems fine and (assuming it has the right effect for users) should strip out the dependencies when you don't need them. It looks like you've id'd the right surface!

qmx:
@frankmcsherry I kinda like your idea, but I don't feel that's how people expect features to work in general - if you take a look on a [feature-heavy Cargo.toml](https://github.com/diesel-rs/diesel/blob/master/diesel/Cargo.toml#L42-L69) you're going to see what I mean.

Anyway, can paint this bikeshed any color you like, it's your bikeshed at the end of the day 😂 

Thanks for the crate and let me know what you want me to do!

frankmcsherry:
I say we go with this, and if people complain we rethink it, but it seems simple enough at the moment not to worry about!

PR #151, opened 2018-04-19
notificator: Allow inspection of FrontierNotificator

antiguru:
Adds pending and has_available functions.


antiguru:
I'm not convinced that has_available should actually take `&mut self` and a frontier or simply check if there are unconsumed timestamps. Now as implementing the stateful operator, it seems to just require that there are no available capabilities, not more.

frankmcsherry:
Looks good!

PR #152, opened 2018-04-25
Fix broken example code in mdbook.

qiuxiafei:
As title.


frankmcsherry:
Cool, thank you very much. Sorry for the delay; got caught behind some deadlines and only now coming up for air. :D

PR #156, opened 2018-05-04
dont detach send thread

frankmcsherry:
This PR avoids detaching the communication threads executing `send_loop`, which should ensure that on clean shutdown these threads are able to drain all outbound communication. This is intended to fix defects observed by @utaal, @jonhoo, @ms705.

@utaal observed hangs at shutdown, which lines up with the issue that the send thread perhaps closed before it could transmit the last of its progress messages, say, leaving other processes stalled waiting for anything. In principle they could see the broken socket and react, but they currently ignore this because .. right.

@jonhoo and @ms705 observed panics in the `send_loop` trying to send at a closed socket. I'm less clear about how this happened, as no process should cleanly exit while another process has data for it (under the correctness of the progress tracking protocol). It could have been a symptom of an unclean exit, but the symptom vanish (or was mitigated) by a delay post-computation before exiting the thread.

Perhaps the PR should be broadened to better detect unclean exits and point fingers at the remote connections that seem to be at fault. This fix *should* be an improvement, independently.

*Edit*: Importantly, this PR has not been tested in the wild yet. It is just chilling here and will not be merged without some testing to see if it does deflect the buggy shutdown, and if crashes remain we can try and extend the PR with further fixes.

### DIFF
--- a/communication/Cargo.toml
+++ b/communication/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "timely_communication"
-version = "0.5.0"
+version = "0.5.1"
 authors = ["Frank McSherry <fmcsherry@me.com>"]
 
 description = "Communication layer for timely dataflow"
--- a/communication/src/networking.rs
+++ b/communication/src/networking.rs
@@ -161,7 +161,8 @@ impl<W: Write> BinarySender<W> {
     fn send_loop(&mut self) {
         let mut stash = Vec::new();
 
-        // block until data to recv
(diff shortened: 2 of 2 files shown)


jonhoo:
@frankmcsherry if you can push a new branch to `eintopf` that uses this, I'd be happy to give it a whirl on EC2 using the same infrastructure it panicked on last time!

frankmcsherry:
Thanks @jonhoo! Let me check it out locally first and confirm that we don't see the stalls we were seeing, at which point I'll put something together. I might also swap out the `unwrap()` calls for `expect()` to get a bit better data on what went wrong. :)

PR #158, opened 2018-05-09
Change state_machine operator to process data in-order

antiguru:
Unconditionally stash data in `state_machine` to prevent out-of-order processing of chunks of data for which the frontier is not `less_than` the chunk's time.


frankmcsherry:
Thanks! This looks good (tests pending, ofc).

For the record it looks like there was at least one bug with `state_machine` where its eager processing logic would fire before pending notifications were processed, allowing accumulation out of order when the eagerly processed input records jump the queue in front of the older updates in the notificator. @antiguru's fix switches things around so that the notificator work is retired before we drain the input, ensuring that pending historical work occurs before any eager input processing.

PR #159, opened 2018-05-14
logging examples

frankmcsherry:
This PR adds two examples, `logging_send.rs` and `logging_recv.rs`. They are meant to be examples of how to enable and observe timely logging. They are not obviously the best examples, but they should give a taste for one way to monitor timely computations live.

If you open two shells, navigate to the timely directory, you should be able to type

    cargo run --example logging-recv -- 1

which will start up the logging listener. For .. reasons .. this is happening over a TCP socket (the easiest way for me to get a live binary stream).

In your other shell, you can type

    cargo run --example logging-send -- 1000 10000000

which will start up a massive exchange microbenchmark, repeatedly sending 1,000 records and synchronizing, repeating 10,000,000 times. You should see output that looks like

```
echidnatron% cargo run --example logging-recv -- 1
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/examples/logging-recv 1`
replayed: (1526320988930924002, TimelySetup { index: 0 }, Operates(OperatesEvent { id: 0, addr: [0, 0, 1], name: "Input" }))
replayed: (1526320988931086202, TimelySetup { index: 0 }, Channels(ChannelsEvent { id: 1, scope_addr: [0, 0], source: (1, 0), target: (2, 0) }))
replayed: (1526320988931173446, TimelySetup { index: 0 }, Operates(OperatesEvent { id: 2, addr: [0, 0, 2], name: "Exchange" }))
replayed: (1526320988931250124, TimelySetup { index: 0 }, Channels(ChannelsEvent { id: 3, scope_addr: [0, 0], source: (2, 0), target: (3, 0) }))
replayed: (1526320988931267648, TimelySetup { index: 0 }, Operates(OperatesEvent { id: 4, addr: [0, 0, 3], name: "Probe" }))
replayed: (1526320988931597789, TimelySetup { index: 0 }, CommChannels(CommChannelsEvent { comm_channel: None, comm_channel_kind: Progress }))
replayed: (1526320988931992460, TimelySetup { index: 0 }, Messages(MessagesEvent { is_send: true, channel: 1, comm_channel: None, source: 0, target: 0, seq_no: 0, length: 1000 }))
replayed: (1526320988932141162, TimelySetup { index: 0 }, Messages(MessagesEvent { is_send: true, channel: 1, comm_channel: None, source: 0, target: 0, seq_no: 1, length: 1000 }))
...
```

which should fly by *fast*. Lots is going on. At the beginning it tells you about the dataflow assembly, and then you start to get a stream of all of the scheduling events going on in the system. A lot of these are pretty dull (not much happens in the event).

If you check out `logging_recv.rs`, you'll notice that we've just imported a timely stream, and indeed all of the logging events are a timely stream timestamped with the recorded nanosecond of the worker. They have different types, and you can totally wire up a custom timely computation based on the log events that come past. Probably some filtering is in order!

More details to follow.

### DIFF
--- a/examples/logging-recv.rs
+++ b/examples/logging-recv.rs
@@ -1,0 +1,32 @@
+extern crate timely;
+
+use std::net::TcpListener;
+use timely::dataflow::operators::Inspect;
+use timely::dataflow::operators::capture::{EventReader, Replay};
+use timely::progress::nested::product::Product;
+use timely::progress::timestamp::RootTimestamp;
+use timely::logging::{TimelySetup, TimelyEvent};
+
+fn main() {
+    timely::execute_from_args(std::env::args(), |worker| {
+
+        let source_peers = std::env::args().nth(1).unwrap().parse::<usize>().unwrap();
+
+        // create replayers from disjoint partition of source worker identifiers.
+        let replayers =
+        (0 .. source_peers)
+            .filter(|i| i % worker.peers() == worker.index())
+            .map(|i| TcpListener::bind(format!("127.0.0.1:{}", 8000 + i)).unwrap())
+            .collect::<Vec<_>>()
+            .into_iter()
+            .map(|l| l.incoming().next().unwrap().unwrap())
+            .map(|r| EventReader::<Product<RootTimestamp, u64>,(u64, TimelySetup, TimelyEvent),_>::new(r))
+            .collect::<Vec<_>>();
+
+        worker.dataflow(|scope| {
+            replayers
+                .replay_into(scope)
+                .inspect(|x| println!("replayed: {:?}", x));
+        })
+    }).unwrap(); // asserts error-free execution
+}
--- a/examples/logging-send.rs
+++ b/examples/logging-send.rs
@@ -1,0 +1,66 @@
+extern crate timely;
+
+use timely::dataflow::InputHandle;
+use timely::dataflow::operators::{Input, Exchange, Probe};
+
+use timely::dataflow::operators::capture::EventWriter;
+
+fn main() {
+    // initializes and runs a timely dataflow.
+
+    use timely::logging::{LoggerConfig, EventPusherTee};
+    let logger_config = LoggerConfig::new(
+        |setup| {
+            use std::net::TcpStream;
+            let addr = format!("127.0.0.1:{}", 8000 + setup.index);
+            let send = TcpStream::connect(addr).unwrap();
+            EventWriter::new(send)
+        },
+        |_setup| {
+            // No support for communication threads in this example.
+            unimplemented!();
+            use std::net::TcpStream;
+            let addr = format!("127.0.0.1:{}", 8001);
+            let send = TcpStream::connect(addr).unwrap();
+            EventWriter::new(send)
+        }
+    );
+
+    timely::execute_from_args_logging(std::env::args(), logger_config, |worker| {
+
+        let batch = std::env::args().nth(1).unwrap().parse::<usize>().unwrap();
+        let rounds = std::env::args().nth(2).unwrap().parse::<usize>().unwrap();
+        let mut input = InputHandle::new();
+
+        // create a new input, exchange data, and inspect its output
+        let probe = worker.dataflow(|scope|
+            scope
+                .input_from(&mut input)
+                .exchange(|&x| x as u64)
+                .probe()
+        );
+
+
+        let timer = std::time::Instant::now();
+
+        for round in 0 .. rounds {
+
+            for i in 0 .. batch {
+                input.send(i);
+            }
+            input.advance_to(round);
+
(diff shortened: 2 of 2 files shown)


lydia-gu:
Hello Frank,

Not sure whether it's proper to ask here. 

We have a study group investigating timely, and I encountered some difficulties when trying to build a capture_send - capture_recv pair using input data in JSON string format. I will be able to see the input string coming into the worker.dataflow. But the capture_recv seems not capturing anything.

Currently the logging code is also hard for me to understand. Appreciate if you could shed some light here. Thank you very much!

```
extern crate timely;

use std::net::TcpStream;
use timely::dataflow::operators::ToStream;
use timely::dataflow::operators::Inspect;
use timely::dataflow::operators::capture::{Capture, EventWriter};

use timely::dataflow::operators::Input;
use timely::dataflow::operators::Probe;
use std::fs::File;
use std::io::{BufReader,BufRead};
use timely::dataflow::*;

fn main() {
    timely::execute_from_args(std::env::args(), |worker| {

        let addr = format!("127.0.0.1:{}", 8000 + worker.index());
        let send = TcpStream::connect(addr).unwrap();

        let index = worker.index();

        let probe = worker.dataflow(|scope|
            scope
                .input_from(&mut input)
                .inspect(move |x| println!("worker {}:\thello {}", index, x))
                .probe()
        );

        let v:Vec<u64> = vec![1526320988930924002, 1526320988931086202, 1526320988931173446];
        let col1 = vec![(0, 0, 1),(1, 0, 1)];


        worker.dataflow::<u64,_,_>(|scope|{
            //v.to_stream(scope).capture_into(EventWriter::new(send))
            let stream1 = input.to_stream(scope);
            stream1.inspect(|x| println!("test: {}", x));
        });



        let file = File::open("test_json.txt").unwrap();
        let mut round = 0;
        for line in BufReader::new(file).lines() {
            input.send(line.unwrap().to_owned());
            input.advance_to(round + 1);
            while probe.less_than(input.time()) {
                worker.step();
            }
        }

    }).unwrap();
}
```

Here is the print output:
```
worker 0:       hello {'id':1,'time':123456789,'qos_id':1}
test: {'id':1,'time':123456789,'qos_id':1}
worker 0:       hello {'id':2,'time':123456790,'qos_id':2}
worker 0:       hello {'id':3,'time':123456791,'qos_id':3}
test: {'id':2,'time':123456790,'qos_id':2}
test: {'id':3,'time':123456791,'qos_id':3}
```

frankmcsherry:
Hi! Happy to help.

This PR is fine, but you are also welcome to open an issue if you like (other people will be more likely to notice it there).

I have to admit, I'm not sure I understand the problem you are seeing. I can believe that there is a problem (and that logging is hard to understand; I'm starting to use it now and .. right. I agree! :).

The code you've written above seems to be performing correctly in terms of the output produced. Did you expect it to produce something else? Or, is there a similar program (with the `capture_into` uncommented) that does something unpleasant when combined with the `capture_recv.rs` side of the socket?

lydia-gu:
Thank you for the prompt help.
Yes, what I tried to do for the next step is to `send` from the `capture_send` to `capture_recv`, using my changed code of `capture_send`:
```
worker.dataflow::<u64,_,_>(|scope|{
            let stream1 = input.to_stream(scope);
            stream1.capture_into(EventWriter::new(send));         
        });
```

However, the `capture_recv` did not return the `replayed` as expected. I just use the `./example/capture_recv` from timely-dataflow repo.

```
$ cargo run --example capture_recv -- 1 w3
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/examples/capture_recv 1 w3`
Illegal instruction: 4
```

frankmcsherry:
Aha! I suspect the problem is that `capture_recv.rs` is attempting to replay the data as `u64` data. The `capture_recv.rs` program has no knowledge about what *type* of data you have serialized at it, and you need to explain that it is (probably) `String`. Unmodified, the code is matched to the previous `capture_send.rs` example, which uses `u64`s.

The line you want to change is

https://github.com/frankmcsherry/timely-dataflow/blob/master/examples/capture_recv.rs#L20

which tells timely which type of `EventReader` you want to use (specifically, what type it should look for in the data).

lydia-gu:
Cool! It works now. Thanks a lot!
`.map(|r| EventReader::<_,String,_>::new(r))`

BTW what if I want to send over a tuple? Does it mean I will only change the type and it's supposed to work as well?

frankmcsherry:
Yup, you just need to make sure the reader attempts to decode the correct type, and all should work well. The type needs to implement the `Abomonation` trait, which many base types and tuples do (though some do not; we can talk about that more if it becomes a problem).

PR #162, opened 2018-05-23
Update sort for rand 0.5

nieksand:
Rand 0.5 has design improvements but also breaking changes: 
https://www.reddit.com/r/rust/comments/8l95zk/rand_050_released/.

I flipped away from StdRng because of the warning in the Rand source code:
https://github.com/rust-lang-nursery/rand/blob/98b940fb1a3cd9e4f84e5b90e5ac003f5e123387/src/rngs/mod.rs#L86

### DIFF
--- a/sort/Cargo.toml
+++ b/sort/Cargo.toml
@@ -16,7 +16,7 @@ license = "MIT"
 rand="*"
 
 [dev-dependencies]
-rand="0.4"
+rand="0.5"
 
 [profile.release]
 opt-level = 3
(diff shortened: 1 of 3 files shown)


frankmcsherry:
Thanks for the PR, and sorry for the delay. This particular code isn't actually in use at the moment (at least, not by me / timely) so it should probably get a once over to figure out what it needs in terms of random number generation. It may be that the data generation for the profiling doesn't need to track major version shifts in RNG. I'll try to get up to speed on the changes, though. Thank you for calling them out!

nieksand:
Fair enough.  I'll close this to reduce clutter.

PR #163, opened 2018-05-24
Add/AddAssign for Product inner

antiguru:
Add convenience `Add`/`AddAssign` implementations for the `Product` type. The implementation can be used to directly increment the inner dimension.

An alternative with less ambiguity would be to restrict `TO` to `RootTimestamp`.


antiguru:
Functionality doesn't seem to be very important, hence closing the PR.

PR #170, opened 2018-06-08
Remove travis-cargo from .travis.yml, publish mdbook on gh-pages

utaal:


frankmcsherry:
Would normally write "LGTM", but no clue here. Merging anyhow!

Issue #165, opened 2018-05-30
Looping In Timely Dataflow


aaj24:
```rust
let probe = worker.dataflow(|scope|{
                let (handle, cycle) = scope.loop_variable(loopNum, 1);
                scope.input_from(&mut input).exchange(|x| *x)
                                            .map(|x| work(&x, opVec[0]))
                                            .concat(&cycle)
                                            .map(|x| work(&x, opVec[2])) // loop body
                                            .connect_loop(handle);
                cycle.map(|x| (hash(&x), x)).probe(); // last line
});
```

Hello, in the code above I am trying to execute the loop body (commented above) _n_ times using a loop then after I am done with the calculations I want to execute the last line. What is happening when I run it is that it executes the last line with the loop too, so both, loop body and last line, are looped. How can I fix this in such a way that after the whole loop is completed, I take the stream and I continue working on it?

frankmcsherry:
@aaj24 If @skaufhold's tips help, excellent!

Otherwise, the reason that you are seeing the last line executed multiple times is that it is inside the loop "scope" and connected to the stream that re-circulates the data for each iteration. What I suspect you want is to use the `leave()` method which lifts a stream out of a scope and presents it in the outer scope. Specifically, you might want something like `branch` up above, and to then call `leave()` on it. If your condition is based on the timestamp (which contains the round of iteration) then you can write an even simpler `branch` implementation that just consults the appropriate timestamp field to determine where the whole message should go.

I'm happy to help prototype this, if you would like. It sounds like both you and Stefan have some use cases that I haven't had yet (in differential dataflow you want all data from all iterations, rather than just the last).

frankmcsherry:
@skaufhold btw that gist'd operator looks very nice! super sweet that the operator builder api worked out for you. :D

If you want to PR it at the repo that would be totally legit (no pressure to, though). It seems like perhaps having a version like yours, maybe one specialized to times-only, perhaps integrated with the `partition` operator (which does something similar but different) would make sense? Partition is currently a bit of a mess; alternately, if I cleaned it up and it looked more like your code would that be ok? :)

skaufhold:
@frankmcsherry When writing it I actually thought too that something like this may have a place directly in timely dataflow! I'm totally up to making a PR with it once we figure out which variant(s) should be directly in the library. I assume a time-only version would be faster because it could pass the data along internally using `give_content`, is that correct?

[...]

PR #172, opened 2018-06-11
Add Branch and BranchWhen operators

skaufhold:
This adds the branch operator that came up in #165 and another operator called `branch_when` that does almost the same but only concerns itself with the times, not the data itself. They split one stream into two depending on a condition that gets passed as a closure.

I've been using these for a while now and since they're so generic I figured I should contribute them here. My only concern would be the overlap between these and `partition`, but they are definitely much more ergonomic to use if you just want a basic "if/else" type functionality.

Related question: Would it be worth it to introduce an even more generic operator that simply does any 1 input -> 2 output operation, in a similar way to the `binary` operator (just 1 -> 2 instead of 2 -> 1)? If such an operator existed, these two could be rewritten a little shorter than they are now, but I'm not sure if there are many use cases for that kind of thing.

### DIFF
--- a/src/dataflow/operators/branch.rs
+++ b/src/dataflow/operators/branch.rs
@@ -1,0 +1,128 @@
+//! Operators that separate one stream into two streams based on some condition
+
+use dataflow::channels::pact::Pipeline;
+use dataflow::operators::generic::builder_rc::OperatorBuilder;
+use dataflow::{Scope, Stream};
+use Data;
+
+/// Extension trait for `Stream`.
+pub trait Branch<S: Scope, D: Data> {
+    /// Takes one input stream and splits it into two output streams.
+    /// For each record, the supplied closure is called with a reference to
+    /// the data and its time. If it returns true, the record will be sent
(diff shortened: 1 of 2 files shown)


frankmcsherry:
This looks good to me!

Re: your related question, I dunno. It's always a toss up between "how often is this needed" and "how painful is the operator builder". It seems relatively rare, and even more so with things like `Branch`. I guess binary operators are kinda mandatory, but you could always do two outputs by re-using a stream, so they have become somehow less crucial.

I propose for the moment doing nothing more general, and just making sure we have pointers to the source for these operators so that folks can just steal them as appropriate. :)

frankmcsherry:
Re: partition, we'll probably do some sort of culling of operators in the future (there are even more ridiculous operators, like `reclock`). Perhaps at that point we'll have some better understanding about the uses of `partition` vs `branch` and grok whether there should be one implementation and the other methods as tasteful wrappers, or two totally different implementations, etc.

PR #174, opened 2018-06-26
Logging to sockets can be enabled with environment variables

utaal:


frankmcsherry:
Seems sane, and useful!

PR #175, opened 2018-07-09
store and reuse logging streams

frankmcsherry:
This PR changes the behavior of the default logger (based on environment variables) to maintain a HashMap under mutex from stream identifier ("setup") to `TcpStream`. When new loggers are requested, it re-uses the same stream.

This should (and seems to) allow multiple dataflows using the same log processor.

@utaal


frankmcsherry:
Btw, this could in principle use *exactly* the same stream. It would complicate the other end (which would not listen for multiple streams, but would instead need to increment initial capabilities by that number). Probably a bit of an invasive change, but perhaps we may eventually want fewer streams coming out of each process.

PR #178, opened 2018-07-13
fix incorrect is_start value in SerializationEvent

qiuxiafei:
as title.


utaal:
Hi! Thanks for the PR.
I think this may be correct as it was: those lines are the end of a serialization event (time spent performing serialization). I’m on the road and commenting from a phone: I’ll double check once I’m back next week.
/cc @frankmcsherry

frankmcsherry:
It looks like the PR is correct, though I had to double check which was the old version and which one the new. There should be paired `true` and then `false` bits set, and pre-PR they are both true. Probably no rush to accept, so let's await consensus!

frankmcsherry:
Happily merging (sorry for the delay) but we will need to re-integrate logging into the new zero-copy tcp allocator.

This event also has the weird property that although it comes from the communication crate, it is actually performed on the worker thread. Perhaps we should migrate the event over there or re-think how we record things (or deprecate the event, if we think serialization is sufficiently non-intrusive). The *de*-serialization events are quite clearly up in user code, in operator logic, and perhaps we should lift serialization there too, somehow.

PR #180, opened 2018-07-23
Multiple fixes

siddharthasahu:


examples/pingpong.rs
```diff
@@ -15,6 +14,7 @@ fn main() {
                   .to_stream(scope)
                   .concat(&cycle)
                   .exchange(|&x| x)
+                  .inspect(move |x| println!("worker {}:\tvalue {}", index, x))
```
frankmcsherry: I think we can't do this without breaking one of the main use-cases for this example: as a latency microbenchmark. I usually use it to test how long it takes to do e.g. 1,000,000 rounds, which should be in the microsecond-per range, and if we add a `println!` in there that isn't going to work any more.

siddharthasahu: Makes sense. Removed this.

frankmcsherry:
I think these look good, thank you very much! I now realize that there isn't much discipline about what the examples are meant to demonstrate, but for sure the previous `distinct` had the defect that if you copy/pasted it into a larger program it would be wrong.

Issue #134, opened 2018-02-19
Unordered input capabilities can be used on any open input


utaal:
As reported by @frankmcsherry.

Capabilities should be tied to their respective inputs. `unordered_input.session(&Capability)` should not allow capabilities that were provided by other inputs (this can be a runtime check).

frankmcsherry:
Possibly fixed by #187. Worth checking out!

PR #187, opened 2018-09-05
Distinct output capabilities

### LINKED ISSUE #186: Distinguish capabilities for multi-output operators
At the moment, a `Capability<T>` is valid for all outputs of a custom operator, and this is enforced by copying the capability changes into the "held capabilities" for all outputs. However, one is able to describe operators without full connections between inputs and outputs. This lets us capture capabilities from an input that should not be able to reach one of the outputs, and then use the capability for that output. Even if the user does not do this, timely may report that the frontier on that output has regressed, because while the operator reports that messages cannot flow to that output the held capability implies that it could still happen.

One solution is to distinguish capabilities by output port, which should be relatively painless for all operators with only one output. For operators with multiple outputs, this would require acquiring and holding distinguished capabilities, which would be a bit of an ergonomic overhead for operators that previously wanted a capability for all outputs. The "overhead" is valuable for operators with non-standard summaries connecting their inputs and outputs, as they very much want to only hold a capability on a subset of outputs.

frankmcsherry:
This PR adds support for distinguished output capabilities, where each `Capability<T>` is bound to a specific operator output. A `CapabilityRef<T>` is able to be downgraded or retained for specific outputs; for compatibility the methods without the output argument use the first output. The `CapabilityTrait` now provides a method that validates the capability for use with an output, which performs `Rc::ptr_eq` on the buffers they would be updating (and in the case of `CapabilityRef` check each possible buffer, corresponding to the outputs).

This adds a bit of ergonomic complexity for folks writing custom operators with multiple outputs, but it is important for anyone who wants to distinguish their output capabilities to enable concurrency that would otherwise be falsely prevented.

This PR does *not* validate the compatibility of the capabilities production according to the operator summary, so a user can still create a capability from an input for an output that they have claimed cannot be reached internally. This would be great to add, but might increase the per-capability reasoning substantially. Future work!

Fixes #186

### DIFF
--- a/src/dataflow/operators/capability.rs
+++ b/src/dataflow/operators/capability.rs
@@ -34,13 +34,20 @@ use progress::ChangeBatch;
 pub trait CapabilityTrait<T: Timestamp> {
     /// The timestamp associated with the capability.
     fn time(&self) -> &T;
+    fn valid_for_output(&self, query_buffer: &Rc<RefCell<ChangeBatch<T>>>) -> bool;
 }
 
 impl<'a, T: Timestamp, C: CapabilityTrait<T>> CapabilityTrait<T> for &'a C {
     fn time(&self) -> &T { (**self).time() }
+    fn valid_for_output(&self, query_buffer: &Rc<RefCell<ChangeBatch<T>>>) -> bool {
+        (**self).valid_for_output(query_buffer)
+    }
 }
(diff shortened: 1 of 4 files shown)


frankmcsherry:
Let's land this! It also appears to fix #134. Once @utaal is back, we can review for non-obvious issues.

PR #188, opened 2018-09-07
Logging revamp

frankmcsherry:
This is a bit of a rework of how logging works in timely dataflow. It is an experiment at the moment, and will need some feedback. 

The main goal is to simplify the use of logging in most cases, pushing some of the complexity on to more complex use cases. For example, logging associated with the communication threads is pushed into building these threads, and while there are sane defaults if you want to change this you need to pick open the thread creation (this shouldn't be any harder than it was before).

The main change is that logging is now user-programmable in the SPMD style of timely dataflow programs, where each worker can programmatically configure the timely logger and loggers for their own custom topics. The timely `Root` type, commonly `worker`, has a method that exposes all available logging endpoints, which can be inserted, removed, and acquired as one likes.

For example, we might start up a timely computation like so, where we register a fairly simple logger of items of type `TimelyEvent` to the `"timely"` topic (which is what the system uses).

```rust
fn main() {
    // initializes and runs a timely dataflow.
    let config = timely::Configuration::from_args(::std::env::args()).unwrap();
    timely::execute(config, |worker| {

        let mut probe = ProbeHandle::new();

        // Register timely worker logging.
        worker.log_register().insert::<TimelyEvent,_>("timely", |_time, data|
            data.iter().for_each(|x| println!("LOG1: {:?}", x))
        );

        // create a new input, exchange data, and inspect its output
        worker.dataflow(|scope| {
            scope
                .input_from(&mut input)
                .exchange(|&x| x as u64)
                .probe_with(&mut probe);
        });
```

This will cause all computation done in the context of this dataflow to be logged with the above logger.

We might then write 

```rust
        // Register timely worker logging.
        worker.log_register().insert::<TimelyEvent,_>("timely", |_time, data|
            data.iter().for_each(|x| println!("LOG2: {:?}", x))
        );

        // create a new input, exchange data, and inspect its output
        worker.dataflow(|scope| {
            scope
                .input_from(&mut input)
                .exchange(|&x| x as u64)
                .probe_with(&mut probe);
        });
```

which will install a different logger and cause the next dataflow to use it for logging instead.

The user is allowed to install arbitrary loggers for arbitrary event types, and can use the `get(name)` method to acquire an installed logger, like so:

```rust
        // Register user-level logging.
        worker.log_register().insert::<(),_>("input", |_time, data|
            for element in data.iter() {
                println!("Round tick at: {:?}", element.0);
            }
        );

        let input_logger = worker.log_register().get::<()>("input").expect("Input logger absent");
```

This creates a new logger that only accepts events of type `()`, which in this case is used to mark round boundaries rather than communicate information. We could also have used a type `String` to report console output for a particular dataflow.

The internal architecture is a bit different than the previous iteration, and it will be good to get @utaal's feedback on it (which design constraints of the previous iteration are ignored here; for each, is that ok?).

The default behavior is meant to be the same as the current implementation, in that it checks for environment variables and uses a TCP socket if interesting variables are set.


frankmcsherry:
This is going to land so's we can move forward! There is now a `CHANGELOG.md` file that should explain several things that have happened (cumulatively), and ideally how to fix things if they have broken. The main thing with this PR is (I think) that `Root` has been rebranded as `Worker` as part of supporting a broader set of things that a worker should do.

PR #195, opened 2018-10-11
Reduce abomonation

frankmcsherry:
This light re-organization removes the requirement that timely dataflow `Data` implement `Abomonation`. Instead, this requirement creeps in when you want to use exchange channels, where we invoke the communication crate's requirements. These could plausibly be reduced (or redefined, with flags) to serde or bincode traits. It isn't clear if they support sizing, but presumably at great cost and with much copying they can be made to work.


frankmcsherry:
This has changed from a "light re-organization" to intending to support swapping out abomonation for bincode, using the "bincode" feature flag. It eagerly deserializes on receipt of data, and just pretends to not know about the "binary" variant of the enum that is either serialized or owned data.

PR #196, opened 2018-10-12
Generalize feedback

frankmcsherry:
This PR changes the interface to `loop_variable` so that the limit and the summary are with respect to the ambient timestamp, not only the one coordinate added by a product combinator. This allows feedback in a greater variety of settings, but adds the complication that one must determine a limit and summary for a more complex type.

The summary should be as simple as `Product::new(Default::default(), summary)`, which isn't especially simple but is at least tractable (and we could simplify it with helper methods).

The limit is harder, because timestamps do not expose a "maximum" method. We would want the limit to be `Product::new(Parent::maximum(), limit)`, but .. right now this is up for debate.

Another option would be to accept a closure determining the limit, which results in a great many `impl Fn(&T)->bool` types in structs (the loop handles), and I don't know the extent to which this is supported yet. Another option is to delete the limit functionality, and ask users to filter their own data if they want to terminate loops early (e.g. with the `branch` method).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ All notable changes to this project will be documented in this file.
 
 - The communication crate now has a `bincode` feature flag which should swing serialization over to use serde's `Serialize` trait. While it seems to work the ergonomics are likely in flux, as the choice is crate-wide and doesn't allow you to pick and choose a la carte.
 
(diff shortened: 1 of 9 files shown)


frankmcsherry:
I've removed the `limit` argument from `loop_variable`, which seems not unreasonable given the existence of the more general `branch_when` operator. The examples were all easy to correct, and the only cases of a non-maximum limit were benchmarks that wanted to artificially terminate the computation.

PR #199, opened 2018-10-19
hang less often

### LINKED ISSUE #198: Race leading to deadlock
Reported by @siddharthasahu 

If you spin up two processes with two worker threads each, running the `hello` example often enough, you eventually hit a deadlock where the two processes do not shut down. It took a while for me to see it, but you do reach something that looks like (the killing happens once you are bored, not before):

```
echidnatron% cargo run --release --example hello -- -n2 -w2 -p0
    Finished release [optimized + debuginfo] target(s) in 0.06s                                                                                               
     Running `target/release/examples/hello -n2 -w2 -p0`
worker 0:	hello 0
worker 1:	hello 1
worker 0:	hello 4
worker 1:	hello 5
worker 0:	hello 8
worker 1:	hello 9
^C
```

and 

```
echidnatron% cargo run --release --example hello -- -n2 -w2 -p1
    Finished release [optimized + debuginfo] target(s) in 0.06s                                                                                               
     Running `target/release/examples/hello -n2 -w2 -p1`
worker 1:	error connecting to worker 0: Connection refused (os error 61); retrying
worker 2:	hello 2
worker 3:	hello 3
worker 2:	hello 6
worker 3:	hello 7
thread 'recv thread 0' panicked at 'assertion failed: read > 0', communication/src/allocator/zero_copy/tcp.rs:64:9
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread 'main' panicked at 'Recv thread panic: Any', libcore/result.rs:945:5
```

No information yet on what is going wrong, but there are for sure a few spots where the data could be getting dropped, buffers could be not flushed, and threads could be not awoken when they should be.

More information once I get this reproducing a bit more consistently, with a read-out of what must have happened / which invariant broke first.

frankmcsherry:
A thorough investigation by a crack team of scientists has concluded that this PR most likely fixes #198. The problem there was a non-deterministic hang on shutdown.

Roughly what was happening is that each send thread has a queue coming from each worker, and when this queue becomes "complete" it is dropped, and once all queues are complete the thread can exit. What constitutes "complete"? We had been using the rule that "if the worker has dropped the send handle", which prevents further sends!

That is pretty smart, but there is something else to check too! See if you can spot it in the diff.

### DIFF
--- a/communication/src/allocator/zero_copy/bytes_exchange.rs
+++ b/communication/src/allocator/zero_copy/bytes_exchange.rs
@@ -78,7 +78,7 @@ impl MergeQueue {
     /// Indicates that all input handles to the queue have dropped.
     pub fn is_complete(&self) -> bool {
         if self.panic.load(Ordering::SeqCst) { panic!("MergeQueue poisoned."); }
-        Arc::strong_count(&self.queue) == 1
+        Arc::strong_count(&self.queue) == 1 && self.queue.lock().expect("Failed to acquire lock").is_empty()
     }
 }
 
--- a/communication/src/allocator/zero_copy/tcp.rs
+++ b/communication/src/allocator/zero_copy/tcp.rs
@@ -135,11 +135,16 @@ pub fn send_loop(
 
         if stash.is_empty() {
             // No evidence of records to read, but sources not yet empty (at start of loop).
-            // We are going to flush our writer (to move buffered data) and wait on a signal.
+            // We are going to flush our writer (to move buffered data), double check on the
+            // sources for emptiness and wait on a signal only if we are sure that there will
+            // still be a signal incoming.
+            //
             // We could get awoken by more data, a channel closing, or spuriously perhaps.
             writer.flush().expect("Failed to flush writer.");
-            signal.wait();
             sources.retain(|source| !source.is_complete());
+            if !sources.is_empty() {
+                signal.wait();
+            }
         }
         else {
             // TODO: Could do scatter/gather write here.


frankmcsherry:
@siddharthasahu @LorenzoMartini

siddharthasahu:
This is not sufficient it seems.

```
% git log --name-status HEAD\^..HEAD
commit c7d1ee771c5f574c2396ad1df2688d8a339a5ae9 (HEAD -> master, origin/master, origin/HEAD, fork/master)
```

```
% git diff                                
diff --git a/communication/src/allocator/zero_copy/bytes_exchange.rs b/communication/src/allocator/zero_copy/bytes_exchange.rs
index 8c6a6f1e..65439e50 100644                       
--- a/communication/src/allocator/zero_copy/bytes_exchange.rs
+++ b/communication/src/allocator/zero_copy/bytes_exchange.rs
@@ -78,7 +78,7 @@ impl MergeQueue {
     /// Indicates that all input handles to the queue have dropped.
     pub fn is_complete(&self) -> bool {
         if self.panic.load(Ordering::SeqCst) { panic!("MergeQueue poisoned."); }
-        Arc::strong_count(&self.queue) == 1
+        Arc::strong_count(&self.queue) == 1 && self.queue.lock().expect("Failed to acquire lock").is_empty()
     }                                    
 }  
``` 

Can still reproduce the bug.

frankmcsherry:
Bah! :)

frankmcsherry:
There was another competing bug, which was certainly a bug but I'm going to be a bit more reserved about my excitement. I've gotten 200 runs in without it hanging, so it may be good.

The problem was that the send thread that should close out when all of the queues on which it waits are empty was waiting on a signal that would never come if while the sender was waiting the last worker sends some records and then exits, dropping the handle. The logic was bad, and the send thread would wait without first checking that there was a non-complete queue, to ensure it would be woken back up.

siddharthasahu:
This might be it!

I have had around 50 runs without seeing the bug again. Once or twice it did take a couple to seconds, but I didn't have to hold my breath for too long.

Looks good on my side.

PR #202, opened 2018-11-06
Use named fields for Cluster in Configuration

utaal:
Arguably easier to read.


frankmcsherry:
Wow. No clue you could do this. :D

PR #203, opened 2018-11-06
 Provide a hook into communication thread spawning

utaal:
This allows customisation of how comm threads are spawned by modifying a `Configuration::Cluster` (similarly to how comm logging is set-up).

`Configuration::Cluster::spawn_fn` has a default
```rust
spawn_fn: Box::new(|index, sender, _, loop_fn| {
    ::std::thread::Builder::new()
        .name(
            if sender { format!("send thread {}", index) }
            else { format!("send thread {}", index) })
        .spawn(move || loop_fn.start())
}),
```

that spawns the thread and starts the respective send/recv loop. The closure receives the comm thread identifier, whether it's a send/recv thread, and the remote target. It can be customised: this enables comm thread pinning:

```rust
Box::new(|index, sender, _, loop_fn| {
    ::std::thread::Builder::new()
        .name(
            if sender { format!("send thread {}", index) }
            else { format!("send thread {}", index) })
        .spawn(move || {
          pin_to_core(base + index);
          loop_fn.start()
        })
}),
```

### DIFF
--- a/communication/src/allocator/zero_copy/initialize.rs
+++ b/communication/src/allocator/zero_copy/initialize.rs
@@ -3,6 +3,7 @@
 use std::sync::Arc;
 use allocator::Process;
 use networking::create_sockets;
+use super::bytes_exchange::{MergeQueue, Signal};
 use super::tcp::{send_loop, recv_loop};
 use super::allocator::{TcpBuilder, new_vector};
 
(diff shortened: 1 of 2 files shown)


utaal:
I tested this with https://github.com/utaal/timely-numa and it appears to work well (the `hello` example there sets a custom `spawn_fn`).

utaal:
And now rebased and ready to merge (if it looks good).

utaal:
Discussed offline: additional refactoring may be worthwhile, so we're holding off this PR for now.

utaal:
I think this is probably out of date at this point. We can re-consider later when communication settles.

PR #205, opened 2018-11-08
Progress hack

frankmcsherry:
This introduces a bit of a hack with respect to progress tracking, strictly optional and enabled by environment variable, that suppresses some volume of progress traffic. It is not meant to be a durable change to the code, but rather to slot into place to help test out to what extent high volumes of progress traffic can be reigned in.

The change is enabled by specifying the environment variable `REACHABILITY_HACK` with value `ZOMG`, which .. should make it clear that this isn't going to stick around (famous last words).

The hack enables logic which will hold back from exchange batches of progress updates for which all changes are demonstrably dominated by other pointstamps, either strictly less pointstamps at the same location, or in the case of messages the existence of an upstream pointstamp with the same timestamp (e.g. a capability held by the source of messages).

This should be generalized and clarified and better tested in the future, but at the moment we only maintain partial information about the state of progress tracking, and these are the only optimizations I could figure that we should be doing.

### DIFF
--- a/examples/realtime.rs
+++ b/examples/realtime.rs
@@ -1,7 +1,7 @@
 extern crate timely;
 
 use timely::dataflow::{InputHandle, ProbeHandle};
-use timely::dataflow::operators::{Input, Filter, Probe};
+use timely::dataflow::operators::{Input, Exchange, Probe};
 
 fn main() {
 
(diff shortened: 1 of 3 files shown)


frankmcsherry:
Closing in favor of #228, which ports this logic to the event-driven scheduler.

PR #214, opened 2018-12-31
Fix second example in README.md

cuongdo:
The second example doesn't compile. The fix is to sync this example with
examples/hello.rs.

---
This change is [Reviewable](https://reviewable.io/reviews/frankmcsherry/timely-dataflow/214)


frankmcsherry:
Ah, thank you! Keeping this file in line is always a pain >.<

I suspect the CI will be fine, as the problem is mainly that CI doesn't test anything in this file. :D

PR #215, opened 2019-01-02
Fix sequence::Sequencer to work with event_driven

comnik:
'Twas not using the activator yet, 'tis now.

### DIFF
--- a/src/synchronization/sequence.rs
+++ b/src/synchronization/sequence.rs
@@ -6,6 +6,7 @@ use std::time::{Instant, Duration};
 use std::collections::VecDeque;
 
 use ::{communication::Allocate, ExchangeData};
+use ::scheduling::Scheduler;
 use worker::Worker;
 use dataflow::channels::pact::Exchange;
 use dataflow::operators::generic::operator::source;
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Cool, thank you!

What do you think about the possibility of activating the operator only when someone pushes a command (i.e., wrapping one of the shared queues so that it has an activator to ping on push)?

comnik:
That sounds like it would be a clear improvement that one could have reasonably expected me to think of from the beginning.

Is it safe to schedule source operators on only a subset of workers in all cases (I don't see why it wouldn't, but still)?

I will take a stab at this tomorrow.

frankmcsherry:
It should be safe to activate only a subset of workers (in the same way that in the normal course of execution, perhaps varying subsets of workers are active for different dataflow operators). Though, it's a good point about sorting out how to communicate what if any requirements there are.

Though actually, I take back the suggestion that this is a good idea, because Sequencer specifically needs all operators firing with high frequency because it uses the timer to gauge progress, and if operators cease advancing their own capabilities, although others might receive a record they may not actually tick forward their clocks. 

One could imagine a more sophisticated design, in which the source operator is activated and produces new outputs only when a user requests, and then the broadcast actions act as a challenge to other operators who must now activate until their source timestamp exceeds the greatest timestamp the second operator has received. This would let the operators go quiet again until a new requests shows up. The operators would lose the ability to confirm that nothing is scheduled for a particular time (they could request this with a no-op event bearing a timestamp), but everyone should still see the same sequence.

comnik:
So I *think* 2d6527a04c39de70c743484d2e75f2c8ffd01c83 does what you said. It doesn't seem to save much when running the sequence.rs example. But all workers still make progress when one of them isn't pushing anything.

frankmcsherry:
This looks roughly like what I had in mind (with more detail, because I had forgotten about the `Rc` sharing and weak upgrades and all of that). Yeah, if you have anything to tidy up, go for it! I think you are currently the main consumer / tester of this, so for sure yell if it is doing something surprising for you.

Also, it isn't exactly meant to speed things up, so much as allow the workers to enter a quiescent state where no one is activated. This is near-future work, but ideally workers who have no work and are promised that they will be awoken by any incoming communication can just go to sleep (if the user permits). This would be an important fix to support that.

PR #220, opened 2019-01-22
Fix the target_ns calculation (in openloop)

utaal:
This uses `leading_zeros` instead of `next_power_of_two` and makes
sure that it always rounds to at least 1024ns.


frankmcsherry:
Looks good! Sorry for the delay in checking this out.

PR #224, opened 2019-02-04
Sequencer: order by sequence, use strong references

frankmcsherry:
This PR introduces to `Sequencer` worker-local sequence numbers to ensure that the worker-local sequence is reflected in the output. Previously, any batched elements would be introduced at the same time and the sequencer would sort them to ensure a consistent order. Now, each worker tags each item with its index and an increasing counter, and the total order is by `(time, worker, counter)`.

The PR also changes the activator handles to be strong references rather than weak references, to avoid crashes in the case that the `Sequencer` is dropped before it is drained. This can happen for workers that only plan to react in-dataflow to sequenced messages, and just need to participate in the sequencing (activating themselves to confirm receipt of messages).

### DIFF
--- a/src/synchronization/sequence.rs
+++ b/src/synchronization/sequence.rs
@@ -59,6 +59,35 @@ impl<T: Ord+ExchangeData> Sequencer<T> {
     /// The `timer` instant is used to synchronize the workers, who use this
     /// elapsed time as their timestamp. Elements are ordered by this time,
     /// and cannot be made visible until all workers have reached the time.
+    ///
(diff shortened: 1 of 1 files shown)


frankmcsherry:
@comnik Take a peek and see if this seems sane to you. Although "breaking" it should mostly be repairing things, but let me know if it works for you, and if the intended behavior seems sane for your cases (and generally).

comnik:
Great, thanks! Testing with 3DF now.

PR #228, opened 2019-02-11
Restrain progress messages

frankmcsherry:
This PR ports the logic from #205 to the new event-driven scheduler logic. Most of the logic is unchanged or simplified, due to the state maintained by the new reachability tracker.

One important change is that the default behavior is now with the progress accumulation optimization *enabled*, rather than disabled as it is in #205. To return to the old behavior, where progress messages are eagerly exchanged rather than exchanged only when justified, set the environment variable

    DEFAULT_PROGRESS_MODE = EAGER

This should cause all progress messages to be exchanged as soon as they are available, even if they cannot obviously improve the progress of the computation.

The down side for non-eager progress message exchange is the potential increase of the critical path for latency sensitive, progress bound computations. Where one `worker.step()` could have circulated all outstanding progress messages, we may need multiple rounds in order to inform participants that they should now send their updates. 

It remains an open question how painful this is, but the alternative (eager progress messages) has a known downside of many orders of magnitude increase in the volume of progress traffic. Ideally, some experimentation will begin to surface any problems, and we can look at ways to fix things further.

### DIFF
--- a/src/progress/nested/reachability_neu.rs
+++ b/src/progress/nested/reachability_neu.rs
@@ -292,10 +292,19 @@ impl<T: Timestamp> PortInformation<T> {
             output_summaries: Vec::new(),
         }
     }
+    /// True if updates at this pointstamp uniquely block progress.
+    ///
+    /// This method returns true if the currently maintained pointstamp
+    /// counts are such that zeroing out outstanding updates at *this*
+    /// pointstamp would change the frontiers at this operator. When the
+    /// method returns false it means that, temporarily at least, there
+    /// are outstanding pointstamp updates that are strictly less than
+    /// this pointstamp.
     #[inline(always)]
     fn is_global(&self, time: &T) -> bool {
-        self.pointstamps.count_for(time) > 0 &&
-        self.implications.count_for(time) == 1
+        let dominated = self.implications.frontier().iter().any(|t| t.less_than(time));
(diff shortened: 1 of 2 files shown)


comnik:
Is there a reason why this is configured via an environment variable and not as a compile-time feature? 

And does either mode affect running in `Configuration::Thread`?

Thanks!

frankmcsherry:
Good questions both. 

1. The environment variable approach lets us better test with and without, to relate the performance of two configurations. The test is not expensive (the variable is read once per scope, and the boolean bound internally), and the compiled code are not much different. Perhaps it should be added to per-scope configuration parameters, but I think at least for experimentation the environment variable approach makes things a bit easier to test, diagnose, and debug.

2. Good question. I think "mostly no". There might be some corner cases where it makes an interesting change. I think I could write such a demonstration (e.g. a scope whose inputs don't advance, but which contains a flatmap that slowly plays out its data; would accumulate updates until the scope input advances). If it does have a noticeable impact something is probably mis-understood on my part, though.

Not planning on landing this without further testing, and perhaps without swapping the defaults back to eager to avoid surprises for folks not paying attention to our public test schedule.

comnik:
Ok, thanks for clearing things up :)

frankmcsherry:
Merged with default changed back to old behavior. Switching default will be a more interesting choice that should call for a bit of experimentation.

PR #235, opened 2019-02-19
Adding data to a `ChangeBatch` will maintain internal bounds

### LINKED ISSUE #234: ChangeBatch accumulates updates without compaction for non-notified operators
Recent changes only call into operators if there is actual work for them. If only the frontier advances but no new data arrives, they will not be notified. However, their frontier changes will only be compacted once they are notified. This can lead to rapidly growing change batches and can cause memory exhaustion.

antiguru:
A `ChangeBatch` absorbs progress updates and previously only compacted
its data once it was read. With this change, `ChangeBatch` will maintain
its internal bounds and also compact when pushing data, but only if
necessary.

This fixes #234.

### DIFF
--- a/src/progress/change_batch.rs
+++ b/src/progress/change_batch.rs
@@ -76,6 +76,7 @@ impl<T:Ord> ChangeBatch<T> {
     #[inline]
     pub fn update(&mut self, item: T, value: i64) {
         self.updates.push((item, value));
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Looks good to me. Thank you very much for the report and fix!

PR #238, opened 2019-02-22
Upgrade to the Rust 2018 edition

benesch:
There were three changes in the 2018 edition of Rust that required
changes:

  1. Removal of the `try!` macro. (Or, rather, the fact that `try` is
     now a keyword means that using the `try!` macro requires using an
     ugly raw identifier, i.e., `r#try!`.) Use the `?` operator instead.

  2. Handling of paths in `use` declarations. Referring to other modules
     in the same crate now requires an explicit `crate::` prefix.

  3. Removal of anonymous trait parameters. All trait parameters now
     require a name. Simply use the name `_` for trait parameters that
     were previously anonymous.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ name = "timely"
 version = "0.8.0"
 authors = ["Frank McSherry <fmcsherry@me.com>"]
 readme = "README.md"
+edition = "2018"
 
 description = "A low-latency data-parallel dataflow system in Rust"
 
--- a/communication/examples/hello.rs
+++ b/communication/examples/hello.rs
@@ -1,5 +1,3 @@
-extern crate timely_communication;
-
 use std::ops::Deref;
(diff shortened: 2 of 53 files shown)


frankmcsherry:
Thanks very much! Let me do a bit of reading on all of this, and try and get my head around any potential negative implications for users (e.g. do people *using* timely all need a Rust version bump now, or will they effectively need one to stay sane).

I'm fairly positive on the "let's move forward" aspect though!

benesch:
> e.g. do people _using_ timely all need a Rust version bump now, or will they effectively need one to stay sane

So they'll need to be using a version of rustc that supports the `crate::` syntax and `?` operator. I'm not sure exactly what version that is. But they don't need to upgrade their own crates to the 2018 edition. 2015 edition crates and 2018 edition crates can interact seamlessly.

frankmcsherry:
Looks good to me!

PR #241, opened 2019-03-01
Shutdown logging

frankmcsherry:
This WIP PR adds logging related to operator and dataflow shutdown. Right now the two are unified, as the creation of a dataflow is also logged as an `OperatesEvent`. Perhaps they could be broken apart to help out folks, but for the moment no.

cc @comnik


frankmcsherry:
This PR is not especially well tested (i.e. not confirmed that we capture all shutdown events) but merging it is probably the right way forward to testing (it should not break existing code, and may only be incomplete for uses of the new code).

comnik:
I can at least verify now that shutdown events are issued for the expected set of operates whenever I use the shiny new arrangement shutdown buttons. Satisfying to see :)

PR #242, opened 2019-03-05
Shush some dead code warnings

benesch:
The code that uses the count field is currently commented out. Comment
out the count field too, for now, to shush some dead code warnings.


frankmcsherry:
Thank you. I've create issue https://github.com/TimelyDataflow/timely-dataflow/issues/243 to track the reason these counters exist, even though that reason is currently disabled for liveness reasons.

PR #245, opened 2019-03-06
Include a magic number in the TCP stream handshake

benesch:
Higher layers may want to multiplex Timely traffic on a port that is also used for other protocols (e.g., HTTP). Sending a magic number immediately after establishing the stream makes it easy to sniff out Timely traffic.

### DIFF
--- a/communication/src/networking.rs
+++ b/communication/src/networking.rs
@@ -78,17 +78,14 @@ pub fn create_sockets(addresses: Vec<String>, my_index: usize, noisy: bool) -> R
 
 /// Result contains connections [0, my_index - 1].
 pub fn start_connections(addresses: Arc<Vec<String>>, my_index: usize, noisy: bool) -> Result<Vec<Option<TcpStream>>> {
-    let mut results: Vec<_> = (0..my_index).map(|_| None).collect();
-    for index in 0..my_index {
-        let mut connected = false;
-        while !connected {
-            match TcpStream::connect(&addresses[index][..]) {
+    let results = addresses.iter().take(my_index).enumerate().map(|(index, address)| {
+        loop {
+            match TcpStream::connect(address) {
                 Ok(mut stream) => {
                     stream.set_nodelay(true).expect("set_nodelay call failed");
                     unsafe { encode(&(my_index as u64), &mut stream) }.expect("failed to encode/send worker index");
-                    results[index as usize] = Some(stream);
                     if noisy { println!("worker {}:\tconnection to worker {}", my_index, index); }
-                    connected = true;
+                    break Some(stream);
                 },
                 Err(error) => {
                     println!("worker {}:\terror connecting to worker {}: {}; retrying", my_index, index, error);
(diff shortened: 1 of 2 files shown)


frankmcsherry:
In principle each timely instance should be talking to the same binaries, which we could also try to check. I wonder if it makes sense to try and generate a magic number at build time, to confirm that the binaries are talking to similar instances of themselves... ?

benesch:
> In principle each timely instance should be talking to the same binaries, which we could also try to check. I wonder if it makes sense to try and generate a magic number at build time, to confirm that the binaries are talking to similar instances of themselves... ?

Indeed it would! I was planning to do that at a higher layer of the stack (i.e., Material), since folks often frown upon doing anything too magical in a library, and Material will need to do its own version check anyway. But there's no harm in the redundancy, if you think its worth doing in Timely too.

Did you mean generating a magic number in build time in addition to or instead of the permafixed magic number? IMO its worth having both. The permafixed magic number reduces the burden on the TCP multiplexer, since it can do a blind equality comparison on the first u64 in the stream. By contrast, the PostgreSQL wire protocol, annoyingly starts a connection with one of five possible versions, with the possibility of expansion at any point; HTTP starts with one of "GET"/"HEAD"/"OPTIONS"/"POST"/"PUT'/.... 

You could argue that perhaps we shouldn't be multiplexing multiple protocols over the same port—that *is* what ports are for, after all—and you would be right, but experience has shown that when one application needs to speak multiple protocols, it is much easier on everyone involved if you can stuff all the protocols into one port.

communication/src/networking.rs
```diff
@@ -109,8 +113,13 @@ pub fn await_connections(addresses: Arc<Vec<String>>, my_index: usize, noisy: bo
...
-        let mut buffer = [0u8;8];
-        stream.read_exact(&mut buffer).expect("failed to read worker index");
+        let mut buffer = [0u8;16];
+        stream.read_exact(&mut buffer)?;
+        let (magic, mut buffer) = unsafe { decode::<u64>(&mut buffer) }.expect("failed to decode magic");
+        if magic != &HANDSHAKE_MAGIC {
+            return Err(io::Error::new(io::ErrorKind::InvalidData,
+                "received incorrect timely handshake"));
+        }
         let identifier = unsafe { decode::<u64>(&mut buffer) }.expect("failed to decode worker index").0.clone() as usize;
```
frankmcsherry: This isn't going to read out the correct identifier, will it?  It seems like it will just read out the magic number again.

benesch: No, it works! `buffer` is advanced (i.e., reassigned to the remaining bytes that abomonation returns) on L118.

frankmcsherry: You are entirely correct! Something feels icky, but we can tidy that up in post. :D

frankmcsherry:
This looks good to me, modulo maybe wanting to tweak things in the future (e.g. per-build magic number in there too, to make sure struct layouts and `usize` sizes are the same; that sort of thing.

benesch:
Yes, please feel free to modify as you see fit! All that matters to me is committing to some backwards-compatible scheme before Materialize ships.

PR #247, opened 2019-03-11
Replace #[inline(always)] with #[inline].

rw:
As a maintainer of another high-performance Rust project (Rust FlatBuffers), I've learned that forcing the compiler to inline can be problematic. The Rust toolchain (LLVM in particular) can make better decisions than we can (usually) when deciding when and what to inline. This PR switches all `#[inline(always)]` attributes to `#[inline]`, in accordance with Rust best practices.

See also https://github.com/TimelyDataflow/differential-dataflow/pull/157


frankmcsherry:
I think going with this makes sense. Perhaps some performance regressions, but at least this gives us a chance to shake them out and explain each of the mandatory inlinings.

PR #250, opened 2019-03-18
kafkaesque: upgrade to the latest librdkafka

benesch:


frankmcsherry:
Looks good. Librdkafka just went 1.0, so there may be a bit of churn in the future, but always good to update; thank you!

PR #251, opened 2019-03-19
Upgrade to Rust 2018, pt. 2

benesch:
When I upgraded the main Timely crate to Rust 2018 in #238, I failed to notice the five subcrates that were still on Rust 2015. Upgrade those too. Also reorg to use Cargo workspaces, to get access to the the fancy new multi-crate tools built into Cargo. Also also attempt to make CI faster.


frankmcsherry:
This all seems good; the workspaces thing weirds me out a bit with the `Cargo.toml` in the `src/` directory. Perhaps that is all natural and sane and things, but still a bit weird. Would it make more sense to shim another directory in place to contain the `Cargo.toml` and `src/` (and any examples) like in the other projects?

benesch:
Yeah, it's definitely more idiomatic to introduce a `timely` directory at the top level and move {tests, examples, src} inside of it. I was just trying to minimize code movement for ease of review, but if you're on board, I'm happy to make the change!

benesch:
Done! GitHub's rename detection support actually worked perfectly here, so there's not even much diff noise.

frankmcsherry:
Okies. Let's try this out. :D

PR #252, opened 2019-03-19
Add GenericSequencer, ThreadSequencer

comnik:
The idea here is to have a more generic sequencer that can drop down to specialized implementations, based on the number of workers (similar to the communication allocators). 

The immediate need was to drop down to a glorified `VecDeque` in single worker mode, such that we can easily run 3DF (and maybe interactive as well?) in single-threaded, blocking mode, without hogging the CPU.

When running with blocking I/O we would otherwise have to be very careful to always reconcile the sequencing dataflow fully, otherwise we might stall.

If you think this is a good idea, @frankmcsherry, I can clean up and add a test. Otherwise I can just keep this within declarative.

edit: sorry I `cargo fmt`-ed a bit too carelessly there.

### DIFF
--- a/timely/examples/sequence.rs
+++ b/timely/examples/sequence.rs
@@ -1,16 +1,15 @@
 extern crate timely;
 
 use std::time::{Instant, Duration};
-use std::collections::VecDeque;
 
 use timely::Configuration;
 use timely::synchronization::Sequencer;
 
 fn main() {
-    timely::execute(Configuration::Process(2), |worker| {
+    timely::execute(Configuration::Process(4), |worker| {
 
         let timer = Instant::now();
-        let mut sequencer = Sequencer::preloaded(worker, Instant::now(), VecDeque::new());
+        let mut sequencer = Sequencer::new(worker, Instant::now());
 
         for round in 0 .. {
             // if worker.index() < 3 {
--- a/timely/examples/sequence.rs
+++ b/timely/examples/sequence.rs
@@ -1,15 +1,16 @@
 extern crate timely;
 
 use std::time::{Instant, Duration};
+use std::collections::VecDeque;
 
 use timely::Configuration;
 use timely::synchronization::Sequencer;
 
 fn main() {
-    timely::execute(Configuration::Process(4), |worker| {
+    timely::execute(Configuration::Process(2), |worker| {
 
         let timer = Instant::now();
-        let mut sequencer = Sequencer::new(worker, Instant::now());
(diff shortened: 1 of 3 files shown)


frankmcsherry:
Probably just within declarative for the moment is best. Ideally I'd like to track down the source of the issue (probably: that sequencer doesn't go quiet even when there is no work?).

There is probably a simpler way to do what you've done, if you are interested:

```rust
pub struct Sequencer<T> {
    activator: Rc<RefCell<Option<CatchupActivator>>>,
    send: Rc<RefCell<VecDeque<T>>>, // proposed items.
    recv: Rc<RefCell<VecDeque<T>>>, // sequenced items.
}
```

You could probably just link `send` and `recv` in the constructor when there is just one worker, not constructing a dataflow at all. Maybe harder to do in declarative, but we could drop that into timely I figure. 

But again, I'd love to know more about the bad thing that sequencer is currently doing. I believe you that there is a thing, but I'm not clear on what it is yet. :D

comnik:
By "link" you mean point them at the same underlying `RefCell<VecDeque<T>>`? I will try that. In general I agree, it doesn't feel like something that shouldn't have to live in core Timely.

I think there is one bad thing that the Declarative server is doing, which is not probing the sequencing dataflow sufficiently. It makes sense to fix that independently of anything else, but I also think it makes sense to avoid constructing the extra dataflow in single-threaded mode.

The busy-waiting in Timely (which I presume is what you're referring to with "the sequencer doesn't go quiet"?) wouldn't be an issue because we would block until user input anyways.

Apart from being useful during development, a single-threaded, embeddable, non-all-CPU-consuming query engine seems pretty doable with what we have already (now that Hector has made expressing complex queries directly in Rust pretty reasonable). I'm thinking of something akin to the now dead [mentat project](https://github.com/mozilla/mentat).

(aside: That is actually how the WebAssembly version of Declarative works right now and I like it a lot, because (for example) it makes it possible to only step dataflows when the browser isn't busy rendering something.)

comnik:
Ok, this seems to work! Got rid of all the new stuff. Only significant change was to make the entire activator construct held in `Sequencer` optional itself (s.t. it still throws if for some reason we `push` before the dataflow has been created).

comnik:
I think with `chill` in master and this can be closed?

frankmcsherry:
I don't need to keep it open myself. If you all have the sense that `chill` achieves the goals this was meant to achieve (calming down workers) then feel free to close. The perf optimization for sequencer is sane to discuss, but .. maybe let's kick that can down the road until we feel the pain of sequencer throughput.

PR #255, opened 2019-03-30
Expose status of activations

eoxxs:
For an event-driven server loop we want to block as long as there is nothing to do.
We don't want to timeout when there are still pending activations and that's the little PR here :) 

We then would not need to probe the sequencer dataflow.

Corresponding stuff over at declarative:
https://github.com/comnik/declarative-dataflow/commit/1a393c4513f93521a43cb8979fe0af266f5eac45

### DIFF
--- a/src/scheduling/activate.rs
+++ b/src/scheduling/activate.rs
@@ -30,6 +30,11 @@ impl Activations {
         self.slices.extend(path);
     }
 
+    /// Are there pending activations?
+    pub fn has_pending(&self) -> bool {
+        (self.bounds.len() - self.clean) > 0
+    }
+
     /// Discards the current active set and presents the next active set.
     pub fn advance(&mut self) {
 


comnik:
Additional context for the "not need to probe the sequencer dataflow": the sequencer sink is already doing all of the relevant logic / activations via the CatchupActivator, and checking pending activations would be more general anyways.

This way, I hope, we can compose dataflows (such as the sequencer) into larger event loops, by letting them be in charge of probing and activating themselves.

frankmcsherry:
Discussed this with Niko and Malte yesterday. While this might be useful, it seems that the intended use may have a race condition (for > 1 workers): if you check this and then sleep (or block) the very next moment a message could arrive and fail (at the moment) to unblock you.

Also, related: this doesn't check the queue coming from the communication plane, so there could be enqueued messages that the worker will learn about in `worker.step()` but which have not yet turned in to activations.

One possible fix that could be short-term appealing for the single-threaded case:

We could add a method `step_or_park(Duration)` which would step if there is any work to do, and if not park itself for some duration. The thread could be awoken with Rust's thread `unpark`, and we could in principle hand out unparking devices to other threads. But, the immediate implementation would have the single-threaded worker behave correctly (as it has total insight into available work) and the multi-worker implementations just step and never park. This allows us to bring the multiworker implementations on-line as that code lands, but you could start using the single worker code sooner.

comnik:
I'm in favour of `step_or_park`.

comnik:
I think this is superseded by #260. @eoxxs?

frankmcsherry:
There is a method `is_empty()` added to activators. It is a bit more conservative, and returns true only when there are neither currently active nor pending activations (because .. that's probably a bug waiting to happen).

comnik:
Oh damn, that one might've been on me. I think David's original version was just like the one in 260 now... Sorry!

eoxxs:
Alright then I'll close this.

PR #261, opened 2019-04-08
Derive impl of Clone for OperatorInfo

utaal:
When writing operators, one may need to pass `operator_info` to various bits of logic, and copying `OperatorInfo` would only happen once during dataflow construction.

@frankmcsherry in no-sync, I construct multiple traces in the same operator - each needs its own copy of `operator_info`.


frankmcsherry:
This seems completely appropriate!

PR #262, opened 2019-04-11
Support u/i128 types

antiguru:
This commit can only be applied once abomination supports u/i128 types.


frankmcsherry:
Looks great (re-ran tests and all green now). Thanks very much!

PR #263, opened 2019-04-17
Add logging::Registry::insert_logger

comnik:
As discussed earlier today, this PR splits `Registry::insert` into `insert_closure` and a lower level `insert_logger`. `insert` defaults to `insert_closure` so this shouldn't be a breaking change.

This makes it possible to create unlogged dataflows without knowing about the logger itself:

``` rust
let timely_logger = worker.log_register().remove("timely");

worker.dataflow(/* unlogged dataflow */);

if let Some(logger) = timely_logger {
  if let Ok(logger) = logger.downcast::<Logger<TimelyEvent>>() {
    worker
      .log_register()
      .insert_logger::<TimelyEvent>("timely", *logger);
    }
}
```

### DIFF
--- a/logging/src/lib.rs
+++ b/logging/src/lib.rs
@@ -15,6 +15,16 @@ pub struct Registry<Id> {
 }
 
 impl<Id: Clone+'static> Registry<Id> {
+    /// Binds a log name to an action on log event batches. See
+    /// `insert_closure`.
+    pub fn insert<T: 'static, F: FnMut(&Duration, &mut Vec<(Duration, Id, T)>)+'static>(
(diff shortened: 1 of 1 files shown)


frankmcsherry:
If `insert` and `insert_closure` do the same thing, perhaps we should just have `insert`?

comnik:
Yeah I kept insert for backwards compatibility, but it's not the most dramatic change. I'll take it out.

frankmcsherry:
Erm, I meant the other way; perhaps we should keep `insert` and not bother with `insert_closure`.

PR #267, opened 2019-04-26
Add Configuration::options

comnik:
This makes it possible for systems using Timely to expose Timely's configuration options to the end user (e.g. print CLI help using Options::usage()).


frankmcsherry:
This seems reasonable. To make sure I understand, the goal here is just to surface the `Options` so that others (people, code) can understand what timely expects as arguments, is that right?

comnik:
Precisely. Simplest use case is to print Timely usage help alongside one's own CLI options.

PR #270, opened 2019-05-22
Fix broken links to simple example

astubbs:
I don't know if you want to change all the references in the book to add the `timely/` prefix or not, but I think the link in the readme should be updated ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ To use timely dataflow, add the following to the dependencies section of your pr
 timely="*"
 ```
 
-This will bring in the [`timely` crate](https://crates.io/crates/timely) from [crates.io](http://crates.io), which should allow you to start writing timely dataflow programs like this one (also available in [examples/simple.rs](https://github.com/timelydataflow/timely-dataflow/blob/master/examples/simple.rs)):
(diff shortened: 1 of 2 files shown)


astubbs:
Let me know which you want and I'll adjust..

frankmcsherry:
Sorry for the delay on this! These look great, and I appreciate the effort of submitting them as a PR!

astubbs:
No worries! But do you want to update all the links in the book? They're all missing the same sub dir.

frankmcsherry:
It occurred to me that you offered to do even more just after I merged, and I didn't want to walk back the good vibes of merging with a request like that. :)

If you are up for link-fixing, that would be amazing. It's not glamorous work, and I can eventually get to it, but if you happen to be grinding through the text with less bleary eyes than I would have that works especially well for me. ;)

astubbs:
Yeah man no worries, I'll take a stab. Just wasn't sure if you
actually wanted to insert the examples prefix into the plain text
links in the book, because it's a bit redundant (i.e. prefix: example
code can be found under the examples folder)...

frankmcsherry:
I think just fixing up the links should be fine. Tbh I'm not sure; I don't have a great read on the expectations of a new reader arriving for the first time; if you have thoughts here I would trust them!

PR #271, opened 2019-05-26
Mutable exchange

frankmcsherry:
This PR changes the closure type used for data exchange from `Fn` to `FnMut`, allowing it to own its own mutable state and preventing it from being called concurrently.

The motivation is that some more sophisticated exchange functions might take e.g. a `Vec<D>` and pick out a subset of elements, and hash the result. This is easiest when the closure owns its own `Vec<D>` that can be populated (or even a `Vec<&D>`, though lifetimes probably prevent that).

### DIFF
--- a/timely/src/dataflow/channels/pact.rs
+++ b/timely/src/dataflow/channels/pact.rs
@@ -43,8 +43,8 @@ impl<T: 'static, D: 'static> ParallelizationContract<T, D> for Pipeline {
 }
 
 /// An exchange between multiple observers by data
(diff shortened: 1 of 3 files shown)


frankmcsherry:
Apparently it also includes some changes to `ParkEvent`...

frankmcsherry:
The PR also raises the question of whether more functions should be `FnMut` rather than `Fn`. For example, the functions for `map` and `filter` are `Fn`, whereas those for `inspect` are `FnMut`.

PR #282, opened 2019-08-05
Fixed typo

namibj:
Please verify the new semantics of the sentence are correct.
I assume they are from reading changelogs.


frankmcsherry:
Your fix is exactly correct!

Unfortunately, I should probably tidy up the text in different ways. That feature has landed, for example. :D But than you very much for the fix!

PR #287, opened 2019-08-21
Add interfaces for activating across thread boundaries

benesch:
Add interfaces that make it easy to send activations across thread
boundaries.

The approach is to give `Activations` an MPSC channel which it checks
for activation requests on every call to `advance`. Other threads can
acquire a handle to the send side of this object via `SyncActivations`
or `SyncActivator`, which are careful to also call unpark on the
associated worker thread for easy interop with `Worker.step_or_park`.

@frankmcsherry I created more interfaces than we actually need just to see how it feels–and I think it feels pretty good! My WIP @MaterializeInc branch only uses `SyncActivator`, so we could potentially inline the `SyncActivations` struct, but the symmetry with `Activator` and `Activations` seemed like it might be nice. We could also expose the underlying `SendError` directly, instead of replacing it with a `SyncActivationError`.

### DIFF
--- a/timely/src/scheduling/activate.rs
+++ b/timely/src/scheduling/activate.rs
@@ -2,15 +2,18 @@
 
 use std::rc::Rc;
 use std::cell::RefCell;
+use std::sync::mpsc::{Sender, Receiver};
+use std::thread::Thread;
 
 /// Allocation-free activation tracker.
-#[derive(Default)]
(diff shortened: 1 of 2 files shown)


frankmcsherry:
Looks nice! Is it ready to go from your point of view?

benesch:
Yessir!

PR #289, opened 2019-08-30
Delayed activations

frankmcsherry:
This PR is a first swing at allowing delayed activations, based on a `std::time::Duration`.

It is not yet tested, but the surrounding methods seem solid. Along with delayed activations, we now have a `replay_core` method that takes an option re-activation period; `Some(Duration::new(0, 0,))` is the standard behavior, any other duration drives a delayed reactivation, and `None` instructs the replayer not to re-activate itself (perhaps someone else will do so).

cc: @comnik, @benesch


comnik:
LGTM 👍 

I think for 3DF I'll have to stick with our own scheduler implementation for now, because we use the same scheduler to trigger events. We don't want to separate those out, to make sure that we do not park while an event is overdue. 

Semi-related: we have experimented with a scheduler implementation that triggers based on a frontier, e.g. to only introduce new inputs whenever a certain amount of previous times have been processed. Is that a pattern that you have tried before?

frankmcsherry:
Sounds good. I'm curious to get my head around which things need to be timely internal, and which things should be a la carte based on what users actually need. In this case, it seemed like the replayer really wanted something like this, which meant that timely needed it, but if there is it comes out that there is a better pattern that has less lock-in (no one wants another tokio) hooray!

The frontier thing: is this like faucet, in spirit? I don't think we've played around with such a thing yet, but it is plausibly interesting!

Potentially related, @benesch wired up rdkafka to surface upcalls and use activators properly; possibly of interest!

comnik:
> Timely Dataflow – Because no one wants another Tokio.

could work!

> Potentially related, @benesch wired up rdkafka to surface upcalls and use activators properly; possibly of interest!

Would definitely be of interest!

> The frontier thing: is this like faucet, in spirit? I don't think we've played around with such a thing yet, but it is plausibly interesting!

Yeah I think so, although I thought Faucet rate limits by handing capabilities from downstream to upstream operators, whereas in our approach the scheduler simply won't schedule operators until the output frontier has made progress. It was mostly a hack and I have no strong opinions on what approach is better / simpler.

---

I think the core issue is the interplay with `step_or_park`, especially `empty_for`. While passing schedulers around makes the operators a bit more verbose, I like how handing out an activation makes for loose coupling with Timely.

The alternative might be a scheduling interface in Timely, any implementation of which could be passed to replay?

frankmcsherry:
I'm going to land this so that I can explore with it, but it does make me think that having scheduling decisions baked into `replay` is a bit weird. I'll chew on this a bit and try to sort out some position.

PR #290, opened 2019-09-03
README.md: Fix description mentioning `new_input`

d-e-s-o:
The README mentions that `new_input` is used for creating the input
stream but that hasn't been the case for a while. Adjust the text to
mention `input_from` which is what the code actually uses.

Bonus: Fix two more typos.


frankmcsherry:
This looks great, thank very much! Apologies for the sloppy work. :D

PR #292, opened 2019-09-04
[WIP] Export TcpBuilder construction

frankmcsherry:
This PR is the first steps in making a la carte construction of channel allocators from TCP connections easier. I expect this to be a bit of an iterative process where we learn about the next painful hunk of code that needs to be copy/paste'd and then split that open too.

cc @benesch


benesch:
Pushed up a commit to https://github.com/MaterializeInc/materialize/pull/424 as proof!

IW00DB:
I do get back with them later and remain quite ignorant. I am not sure what to do and I will not be in that vector today. Would the good fella that I saw today share some advice? I would not want to step on toes. maythesacred@gmail.com

PR #294, opened 2019-09-05
Improve `timely::execute` documentation.

ryzhyk:
Clarify `timely::execute` behavior, namely (see #294):
- the method returns instantly
- dropping the value returned by the method will block until the timely
  computation terminates


frankmcsherry:
These look helpful, thank you!

PR #297, opened 2019-10-09
Update abomonation_derive to 0.4

d-e-s-o:


d-e-s-o:
We actually have to bump to `0.5` now because the owner of `abomonation_derive` decided to publish the most recent commit as a minor version upgrade, not a patch level one.

frankmcsherry:
I'm not 100% clear on what is afoot, other than that there are now (pre-PR) lots of errors spewing out from `abomonation_derive`. Is that what this is about?

d-e-s-o:
Most of the crate eco system moves on to consuming the most recent version of a crate once available (assuming compatibility, of course). `abomonation_derive` was consuming pre-1.0 versions of `proc-macro2`, `syn`, and `quote`, whereas virtually nobody else was. That results in unnecessary compilation overhead because multiple version of a crate are now being compiled. Presumably it also bloats the binary.

So, by moving to `0.5` we avoid that.

Mind sharing what errors you are referring to?

frankmcsherry:
For example in the `cargo doc` region of 

https://travis-ci.org/TimelyDataflow/timely-dataflow/builds/602816414?utm_source=github_status&utm_medium=notification

I'm not sure the link will work, but if not I can copy paste. 

Mostly, I was trying to figure out what the updates to `abomonation_derive` were and if this update was important for functionality, or just to track things. It's the sort of thing that needs to be coordinated, as downstream users usually need to coordinate too, and I didn't want to do that too randomly.

frankmcsherry:
But, I'm not clear on any reason not to do this, so might as well, right?

frankmcsherry:
Thanks very much for noticing the shift (I had no clue) and for keeping dependencies tidy!

d-e-s-o:
Thanks! For what it's worth, I see the same warnings for Travis run [599933814](https://travis-ci.org/TimelyDataflow/timely-dataflow/builds/599933814?utm_source=github_status&utm_medium=notification) which does not include my change.

frankmcsherry:
Oh, I actually meant that *without* the PR I was seeing the warnings, and I was imagining that perhaps the change was meant to address a related issue.

PR #298, opened 2019-10-17
Re-order channel interactions

frankmcsherry:
This PR amends the order of pushing of elements into shared queues to avoid a race in which a thread is first alerted and then shown data. The details are not 100% clear to me at the moment, but we should try it out.

cc @ryzhyk

### DIFF
--- a/communication/src/allocator/counters.rs
+++ b/communication/src/allocator/counters.rs
@@ -91,13 +91,13 @@ impl<T, P: Push<T>> Push<T> for ArcPusher<T, P> {
         // else {
         //     self.count += 1;
         // }
+        self.pusher.push(element);
+
         let _ =
         self.events
             .send((self.index, Event::Pushed(1)));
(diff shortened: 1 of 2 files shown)


ryzhyk:
I think I already tried this. It deadlocks even sooner than the original code :(

frankmcsherry:
I think the re-ordering does fix an issue, even if it ... makes everything else worse. That event counter thing is what prompts scheduling of operators (the `usize` drives a "go read the operator attached to this channel" activation). The counter value is ignored, which 1. perhaps it shouldn't be, and 2. implies that there is a one shot chance to read the message and if it is missed may result in a hang.

frankmcsherry:
The most recent commit ensures that the buzzer is buzzed after the event count is pushed, which is done after the actual message is enqueued. Several bugs here, but it seems like they have chilled out a little bit at this point.

PR #313, opened 2019-12-11
Timestamp::minimum()

frankmcsherry:
This PR introduces an explicit function `minimum()` to use for a minimum timestamp, rather than relying on `Default::default()` which we previously used. This should allow the use of signed integers, where we previously forced the default capability to be zero.

There is fallout in differential, where the use of integer timestamps required that the minimum element of the lattice would need to be zero rather than a small negative number. There is possibly fallout in other projects where folks have used signed integer timestamps, which will no longer start at zero and instead start at a very negative number.


frankmcsherry:
For reasons that I presume relate to incompetence, I picked up some other uncommitted work (deleting deprecated methods, and providing a dataflow dropping method). Both of those seem fine to land also, and to start to try out.

PR #315, opened 2019-12-17
ActivateCapability improvements

benesch:


timely/src/dataflow/operators/capability.rs
```diff
@@ -257,6 +258,55 @@ pub fn mint_ref<'cap, T: Timestamp>(time: &'cap T, internal: Rc<RefCell<Vec<Rc<R
     }
 }
 
+/// Capability that activates on drop.
+#[derive(Clone)]
+pub struct ActivateCapability<T: Timestamp> {
+    pub(crate) capability: Capability<T>,
+    pub(crate) address: Rc<Vec<usize>>,
+    pub(crate) activations: Rc<RefCell<Activations>>,
```
benesch: @frankmcsherry the `pub(crate)`s here are weird, but not sure how to solve.

frankmcsherry: I think we can sort these out in post. It looks like `autoflush_session()` could plausibly just take the `ActivateCapability`.

PR #317, opened 2019-12-19
Add, clone, debug, deref (with target &[Capability<T>]) and from_elem to CapabilitySet

utaal:
`from_elem` allocates a `CapabilitySet` containing a single capability.
This is useful when making a `CapabilitySet` out of the default
capability provided to a new operator (as shown in the example for
`from_elem`).
This is an ergonomics improvement. `from_elem` is the same name used
in `Antichain`:
https://github.com/TimelyDataflow/timely-dataflow/blob/8f0c22b1523784e28d3ac6e5eab30fa6d0a03839/timely/src/progress/frontier.rs#L73-L82

`deref` lets the operator use the capabilities in the set to e.g. open a
session and send messages (again, as shown in the example for
`from_elem`).

### DIFF
--- a/timely/src/dataflow/operators/capability.rs
+++ b/timely/src/dataflow/operators/capability.rs
@@ -308,6 +308,7 @@ impl<T: Timestamp> Drop for ActivateCapability<T> {
 }
 
 /// A set of capabilities, for possibly incomparable times.
+#[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


frankmcsherry:
This looks good to me; thanks very much!

utaal:
I've added a commit that derives `Clone` (which does the obvious thing of cloning the capabilities) and `Debug`.

PR #320, opened 2020-02-11
Make crate paths absolute

malthe:
The relative URLs don't seem to work anymore.


frankmcsherry:
Thanks very much! I would never have noticed, otherwise!

PR #321, opened 2020-03-05
Use timely's logging infrastructure to log Tracker state

saradecova:
Add TrackerEvent which records additions or removals
of capabilities, as well as propagation events when changes
in implications are propagated along the internal connections
and edges of the graph.

Add DebugEvent which records the state of pointstamps,
implications, and worklist of Tracker.

Enabling loggers for these events is done the same way as for
logging::TimelyEvent's

At the moment, we use these loggers for comparative testing between
Isabelle implementation of progress tracking and our Rust implementation.

This [graph](https://github.com/TimelyDataflow/timely-dataflow/files/4293118/barrier.pdf) plots the runtime of the computation with and without the changes, and suggests
that the increase is not significant despite the changes being on the critical path.
The example used is timely/examples/barrier.rs with 10,000,000 samples, run
with four workers. The graph shows complementary cdf for:

1. current master,
2. change with all loggers disabled (almost identical to (1)),
3. original TimelyEvent logger, and
4. original TimelyEvent logger + TrackerEvent logger.

We got similar results when run on other examples.

### DIFF
--- a/timely/src/logging.rs
+++ b/timely/src/logging.rs
@@ -53,6 +53,8 @@ pub struct OperatesEvent {
     pub id: usize,
     /// Sequence of nested scope identifiers indicating the path from the root to this instance.
     pub addr: Vec<usize>,
+    /// Internal summary for every combination of input and output port.
+    pub internal_summaries: Vec<Vec<String>>,
     /// A helpful name.
     pub name: String,
 }
--- a/timely/src/logging.rs
+++ b/timely/src/logging.rs
@@ -47,6 +47,17 @@ impl<T, E, P> Drop for BatchLogger<T, E, P> where P: EventPusher<Duration, (Dura
 }
 
 #[derive(Serialize, Deserialize, Abomonation, Debug, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
+/// The creation of a `Subgraph`.
+pub struct SubgraphEvent {
+    /// Worker-unique identifier for the operator.
+    pub id: usize,
+    /// Sequence of nested scope identifiers indicating the path from the root to this instance.
+    pub addr: Vec<usize>,
+    /// Type of the subscope Timestamp.
+    pub timestamp_type: String,
+}
+
+#[derive(Serialize, Deserialize, Abomonation, Debug, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
 /// The creation of an `Operate` implementor.
 pub struct OperatesEvent {
     /// Worker-unique identifier for the operator.
@@ -223,6 +234,8 @@ impl ParkEvent {
 #[derive(Serialize, Deserialize, Debug, Clone, Abomonation, Hash, Eq, PartialEq, Ord, PartialOrd)]
 /// An event in a timely worker
 pub enum TimelyEvent {
+    /// Subgraph creation.
+    Subgraph(SubgraphEvent),
     /// Operator creation.
     Operates(OperatesEvent),
     /// Channel creation.
@@ -251,6 +264,10 @@ pub enum TimelyEvent {
     Park(ParkEvent),
     /// Unstructured event.
     Text(String),
+}
+
+impl From<SubgraphEvent> for TimelyEvent {
+    fn from(v: SubgraphEvent) -> TimelyEvent { TimelyEvent::Subgraph(v) }
 }
 
 impl From<OperatesEvent> for TimelyEvent {
--- a/timely/src/progress/frontier.rs
+++ b/timely/src/progress/frontier.rs
@@ -317,6 +317,13 @@ impl<T: PartialOrder+Ord+Clone> MutableAntichain<T> {
         self.frontier().less_equal(time)
     }
 
+    /// Clones the vector of updates.
+    /// Only used for debugging purposes.
+    #[inline]
+    pub fn updates(&self) -> Vec<(T, i64)> {
+        self.updates.clone()
+    }
+
     /// Allows a single-element push, but dirties the antichain and prevents inspection until cleaned.
     ///
     /// At the moment inspection is prevented via panic, so best be careful (this should probably be fixed).
--- a/timely/src/progress/frontier.rs
+++ b/timely/src/progress/frontier.rs
@@ -317,13 +317,6 @@ impl<T: PartialOrder+Ord+Clone> MutableAntichain<T> {
         self.frontier().less_equal(time)
     }
 
-    /// Clones the vector of updates.
-    /// Only used for debugging purposes.
-    #[inline]
-    pub fn updates(&self) -> Vec<(T, i64)> {
-        self.updates.clone()
-    }
-
     /// Allows a single-element push, but dirties the antichain and prevents inspection until cleaned.
     ///
     /// At the moment inspection is prevented via panic, so best be careful (this should probably be fixed).
--- a/timely/src/progress/reachability.rs
+++ b/timely/src/progress/reachability.rs
@@ -811,30 +811,44 @@ fn summarize_outputs<T: Timestamp>(
 #[derive(Serialize, Deserialize, Abomonation, Debug, Clone)]
 /// Log event
 pub struct UpdateSourceEvent {
-    worker_id: usize,
-    scope_addr: Vec<usize>,
-    operator: usize,
-    port: usize,
-    timestamp: String,
-    delta: i64,
+    /// Worker index.
+    pub worker_id: usize,
+    /// Sequence of nested scope identifiers indicating the path from the root to this scope.
+    pub scope_addr: Vec<usize>,
+    /// Operator index.
+    pub operator: usize,
+    /// Port index.
+    pub port: usize,
+    /// Timestamp for which capability was added or removed.
+    pub timestamp: String,
+    /// Multiplicity of the capability to added or removed.
+    pub delta: i64,
 }
 
 #[derive(Serialize, Deserialize, Abomonation, Debug, Clone)]
 /// Log event
 pub struct UpdateTargetEvent {
-    worker_id: usize,
-    scope_addr: Vec<usize>,
-    operator: usize,
-    port: usize,
-    timestamp: String,
-    delta: i64,
+    /// Worker index.
+    pub worker_id: usize,
+    /// Sequence of nested scope identifiers indicating the path from the root to this scope.
+    pub scope_addr: Vec<usize>,
+    /// Operator index.
+    pub operator: usize,
+    /// Port index.
+    pub port: usize,
+    /// Timestamp for which capability was added or removed.
+    pub timestamp: String,
+    /// Multiplicity of the capability to be added or removed.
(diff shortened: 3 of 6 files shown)


timely/src/progress/frontier.rs
```diff
@@ -317,6 +317,13 @@ impl<T: PartialOrder+Ord+Clone> MutableAntichain<T> {
         self.frontier().less_equal(time)
     }
 
+    /// Clones the vector of updates.
+    /// Only used for debugging purposes.
```
frankmcsherry: Can you say a bit more about what this is used for in the comments? If this lands, I'll need to support it, and it would help to understand why it is here and under which circumstances it could go away.

timely/src/progress/frontier.rs
```diff
@@ -317,6 +317,13 @@ impl<T: PartialOrder+Ord+Clone> MutableAntichain<T> {
...
+    #[inline]
+    pub fn updates(&self) -> Vec<(T, i64)> {
```
frankmcsherry: Could this return a `&[(T, i64)]` instead, to avoid a mandatory allocation?

timely/src/progress/reachability.rs
```diff
@@ -145,8 +147,9 @@ pub struct Builder<T: Timestamp> {
 impl<T: Timestamp> Builder<T> {
 
     /// Create a new empty topology builder.
-    pub fn new() -> Self {
+    pub fn new(path: Vec<usize>) -> Self {
```
frankmcsherry: I think this is a big abstraction change, in that `reachability.rs` used to be agnostic to the hierarchical nature of names in timely dataflow, and just did reachability tracking in a scope with no additional information. This seems to bake that in now, which .. will have to ponder whether that is a good call or not.

Another option might be to produce another implementation of `reachability.rs`, putting things behind a trait.

Not trying to be difficult, but attempting to minimize the complexity in an already-too-complicated bit of logic.

frankmcsherry:
I have some general questions about the PR!

I think I understand the goal, which is to get out information about the steps that `reachability.rs` takes as it runs. What I'm less clear on is whether this is the best way to do that. I *think* that what's in `reachability.rs` is essentially deterministic once its inputs are specified. So, if we captured the timestamp changes that are supplied as inputs, and the moments at which propagation happens, I think this would be sufficient information to re-execute the code and see what happens, but without asking `reachability.rs` to record this for you.

Would it serve your purposes just as well to instrument the moments and nature of pointstamp updates and propagation?

timely/src/progress/frontier.rs
saradecova: Removed the function as it's no longer needed.

timely/src/progress/frontier.rs
saradecova: Removed the function completely as it's no longer needed.

timely/src/progress/reachability.rs
saradecova: That's a good point. I removed the path argument from from Tracker and Builder constructor. 
However, it is still added to the tracker when the logger is registered. The logging events such as "a capability is added/removed from a location" do not make sense if they cannot be tracked down to a particular location.

Let me know what you think please and whether you would still prefer going your way.

saradecova:
Thank you for the comments! I changed the logging so there is only one additional logger (as apposed to two) logging the minimum information to re-execute the code as you suggested.

utaal:
Hi folks, a couple of more notes from a conversation with @saradecova .

We're using `Debug` to convert timestamps of arbitrary types to `String`s for logging:

https://github.com/TimelyDataflow/timely-dataflow/pull/321/files#diff-d233d1f563e31c39b0ad93a7182fdcb3R493

This seems reasonable (we wouldn't know how to encode the type otherwise), however this makes it harder (impossible) for the consumer to know what type to expect for a certain scope. As an example, to be able to replay the behaviour of the `Tracker` using a trace generated with these events, we need to parse the timestamps (to re-establish the partial order, mainly).

We're wondering if it makes sense to add (or adjust an existing) `TimelyEvent` to output type information for scope timestamps. This way, a consumer can determine how to parse the stringified timestamps (possibly by providing `FromStr` implementations).

One option would be to use `TypeId` ([docs](https://doc.rust-lang.org/std/any/struct.TypeId.html)), which unfortunately hides its internals, and makes it hard (impossible) to parse at the receiver.

I'm still considering options, but @frankmcsherry let us know if you have opinions.

saradecova:
To follow on @utaal and our conversation, we can encode the type in string using `std::any::type_name`.

Moreover, by adding `internal_summaries` to `OperatesEvent` it would be also beneficial to have the timestamp type for users of `TimelyEvent` log-stream.

In pracrise, this could be an event informing us that "A new subscope was created at address `addr` from the root with associated `timestamp_type`:
```
/// The creation of a `Subgraph`.
pub struct SubgraphEvent {
    pub addr: Vec<usize>,
    pub timestamp_type: String,
}
```

utaal:
Also, it may be the right time to address these todo(s) in `ProgressEvent`:

https://github.com/TimelyDataflow/timely-dataflow/blob/06fac10974443cd6654e5c7bd81e1d557f9a780e/timely/src/progress/broadcast.rs#L68-L69

https://github.com/TimelyDataflow/timely-dataflow/blob/06fac10974443cd6654e5c7bd81e1d557f9a780e/timely/src/progress/broadcast.rs#L118-L119

frankmcsherry:
I'm back to looking at this. Very sorry for the delay.

I have several spot comments, and generally think that before landing durably in timely it needs a bit more design work. In particular, 

1. There are some use cases close to but not yet well-served by this (most interesting one from several users: which (operator, timestamp) pairs are on the frontier). I'd like to sort these out, but that doesn't have to block this.

2. The use of `String` for timestamps weirds me out. It also seems to be used for `Antichain`, which .. I can see how this gets you the data you need, but it can't be the best way. :D

3. The change to `OperatesEvent` is substantial; I wonder if it could be normalized out in to another stream?

I still don't have a great read on the requirements here. I apologize if my comments have been confused. My understanding is that you want to be able to extract progress information from the reachability subsystem, and I'm guessing that is to drive your work on progress stuffs. Do you need the fine-grained update information, or just the aggregate information extracted in `propagate_all`? Do you expect the information to be helpful outside of your work (roughly: should this be a branch, or is it valuable to have all timely users have access to this)?

I'm currently trying to reconcile this with requests other information asks about progress tracking, that I think are more about "log the state of the dataflow-wide frontier". It probably relates, and definitely has the same awkwardness around timestamps being generic.

Anyhow, I'm thinking about this now, and trying to understand which things are important to log and which are optional!

frankmcsherry:
I have another ask: is there a qualitative difference for you between logging the progress updates in `update_target` and `update_source` as you do, vs perhaps logging them in `propagate_all` as (or just before) they are drained out of their buffers. The latter has the benefit of being a bit less noisy (we have consolidated the updates and potentially canceled some out), but it would remove visibility in to the first moments at which the tracker knows about some updates (in case that was part of your study).

This looks like a good direction to go to expose information about the state of system progress, but I bet my requirements are not your requirements (e.g. that change above is fine for me, but I don't want to do it if it breaks your reqs).

There are some stray lines added and removed; I can fix these up. Mostly I'm trying to map out "imagine this lands; what 'improvements' should be prohibited?"

timely/src/progress/subgraph.rs
```diff
@@ -165,7 +169,19 @@ where
...
+        tracker.tracker_logger = worker.log_register().get("timely/tracker").map(|logger| (path.clone(), logger));
+
+        // Perhaps log information about the creation of subgraph.
+        if let Some(l) = self.logging.as_mut() {
+            l.log(crate::logging::SubgraphEvent{
+                id: worker.index(),
+                addr: path.clone(),
+                timestamp_type: std::any::type_name::<TInner>().to_string(),
+            });
+        }
```
frankmcsherry, line 184: I'd love to reframe this as a `TrackerEvent` and have it be part of the line just up above (i.e. "tracker came in to existence". I suspect something like `Tracker::install_logger(...)` could do both of those things and wrap up the abstraction well. I'm happy to do that after the fact if that works for you.

timely/src/progress/subgraph.rs
```diff
@@ -524,6 +535,10 @@ where
             }
         }
 
+        // double-check that child 0 (the outside world) is correctly shaped.
+        assert_eq!(self.children[0].outputs, self.inputs());
+        assert_eq!(self.children[0].inputs, self.outputs());
+
```
frankmcsherry, line 541: Can you explain why these moved down 20 lines?

timely/src/worker.rs
```diff
@@ -419,15 +419,18 @@ impl<A: Allocate> Worker<A> {
 
         let mut operator = subscope.into_inner().build(self);
 
+        let (internal_summary, _) = operator.get_internal_summary();
+
```
frankmcsherry, line 423: If at all possible, I'd like to keep this next to the `set_external_summary()` call just to be clear that they are paired. I'm happy to have it hoisted as well.

timely/src/logging.rs
```diff
@@ -46,13 +46,26 @@ impl<T, E, P> Drop for BatchLogger<T, E, P> where P: EventPusher<Duration, (Dura
...
+
 #[derive(Serialize, Deserialize, Abomonation, Debug, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
 /// The creation of an `Operate` implementor.
 pub struct OperatesEvent {
     /// Worker-unique identifier for the operator.
     pub id: usize,
     /// Sequence of nested scope identifiers indicating the path from the root to this instance.
     pub addr: Vec<usize>,
+    /// Internal summary for every combination of input and output port.
+    pub internal_summaries: Vec<Vec<String>>,
```
frankmcsherry, line 68: Can you explain what these are used for? Would it be equally beneficial to have the report from the tracker about its input-to-output summaries? That would leave this event stable and consolidate the timestamp/summary related events to the reachability tracker.

timely/src/progress/reachability.rs
```diff
@@ -597,7 +629,6 @@ impl<T:Timestamp> Tracker<T> {
         //       will discover zero-change times when we first visit them, as no further
         //       changes can be made to them once we complete them.
         while let Some(Reverse((time, location, mut diff))) = self.worklist.pop() {
-
```
frankmcsherry, line 600: random whitespace

timely/src/progress/reachability.rs
```diff
@@ -654,6 +685,7 @@ impl<T:Timestamp> Tracker<T> {
                 };
             }
         }
+
```
frankmcsherry, line 688: random whitespace

frankmcsherry:
Note to self: this would be great pointed at the new logging channel now introduced in #352.

frankmcsherry:
Closing in favor of #375 which borrows heavily from this. If it turns out that there is an urgent need for e.g. logging the topology information, I can certainly make that happen too.

PR #322, opened 2020-03-09
Add PartialOrder impls

frankmcsherry:
This PR adds `PartialOrder` implementations for `Antichain` and `AntichainRef`. The methods `less_equal` and `less_than` clash with the per-time implementations (perhaps those should only be for antichains, which each singleton time is), but the inherent methods take priority in resolution, and this doesn't *appear* to cause any breakage.

The intent here is to allow these types to implement `Lattice` in the differential crate, in the interest of tidying up a bunch of bespoke logic.

cc @LorenzSelv

### DIFF
--- a/timely/src/progress/frontier.rs
+++ b/timely/src/progress/frontier.rs
@@ -138,9 +138,10 @@ impl<T: PartialOrder> Antichain<T> {
     }
 
     /// Returns true if every element of `other` is greater or equal to some element of `self`.
+    #[deprecated(since="0.12.0", note="please use `PartialOrder::less_equal` instead")]
     #[inline]
     pub fn dominates(&self, other: &Antichain<T>) -> bool {
-        other.elements().iter().all(|t2| self.elements().iter().any(|t1| t1.less_equal(t2)))
+        <Self as PartialOrder>::less_equal(self, other)
     }
 
     /// Reveals the elements in the antichain.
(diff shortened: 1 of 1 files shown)


LorenzSelv:
That looks good. Maybe we could call the `dominates` method inside the implementation of `less_equal`, but it's just a minor thing.

Thanks for taking the time to do this.

frankmcsherry:
Oh, that's a good idea. I'd be up for deprecating `dominates` too, with the notes that one should use `PartialOrder::less_equal`. My guess is that it would be great to get everyone to using the partial order on antichains (with a `&[time]` for single times), but that would probably be quite breaking without a lot of socialization. I think the users of `dominates` are mostly in differential.

LorenzSelv:
Yeah that would be even better, up to you

timely/src/progress/frontier.rs
```diff
@@ -157,6 +164,19 @@ impl<T: PartialOrder> Antichain<T> {
     #[inline] pub fn elements(&self) -> &[T] { &self.elements[..] }
 }
 
+impl<T: PartialEq> PartialEq for Antichain<T> {
+    fn eq(&self, other: &Self) -> bool {
```
LorenzSelv: not sure if it works here as I haven't tried, but.. can we just do `self.elements() == other.elements()` ? slice comparison should do the job.

frankmcsherry: It has the defect that it doesn't equate two slices that have the same elements, but in a different order. I think this was one of the reasons that I shied away from the implementation in the first place, but it's probably worth fixing (as otherwise equality testing could return surprising results anyhow).

LorenzSelv: I assumed they were sorted, but it's not the case.. also, how can you sort incomparable elements? ahah nvm, dumb comment

PR #325, opened 2020-04-17
Fix some comments

umanwizard:


frankmcsherry:
These look great, thanks for the catch and sorry for the confusion!

PR #326, opened 2020-05-07
Properly log progress messages

frankmcsherry:
Log progress messages, converting timestamps to strings. This puts more allocation behind each of these messages (from an empty vector to a vector of strings) but there doesn't seem to be much to do about it if we want to see the actual progress contents moving around.

The receive side has the potentially to be much spammier than the send side, as this is a broadcast. Depending on the volume we may want to dial that down and just presume that sent messages are not lost in transit (we would lose the timing information about when they are received though). For the moment, we log lots of things.

cc: @utaal @saradecova

### DIFF
--- a/timely/src/progress/broadcast.rs
+++ b/timely/src/progress/broadcast.rs
@@ -1,7 +1,7 @@
 //! Broadcasts progress information among workers.
 
 use crate::progress::{ChangeBatch, Timestamp};
-use crate::progress::Location;
+use crate::progress::{Location, Port};
 use crate::communication::{Message, Push, Pull};
 use crate::logging::TimelyLogger as Logger;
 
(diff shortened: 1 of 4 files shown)


frankmcsherry:
This doesn't address the same issues as #321, but we should double check whether it relieves any pressure there. There should be only some gap between the receive stream and the inputs to the progress tracker.

frankmcsherry:
Thinking out loud, although these logging events are informative, they aren't exactly sufficient to track the frontier of the system (which would be a desirable goal). For example, operators are often created with an initial capability, which their logic then drops and workers confirm by sending progress messages around. So we see the drops, but not the initial capabilities.

It would be nice to directly log the dataflow frontier in the progress tracker, which is a bit more in line with #321 (though, not something that #321 produces as output). Afaict, #321 produces the raw pointstamp updates in progress tracking, which is possibly more immediately helpful than the progress messages for determining the state of capabilities in the system.

timely/src/progress/broadcast.rs
```diff
@@ -108,16 +125,33 @@ impl<T:Timestamp+Send> Progcaster<T> {
...
-                messages: Vec::new(),
-                internal: Vec::new(),
-            }));
+            self.logging.as_ref().map(|l| {
+
+                let mut messages = Vec::with_capacity(changes.len());
+                let mut internal = Vec::with_capacity(changes.len());
+
+                // TODO: Reconsider `String` type or perhaps re-use allocation.
+                for ((location, time), diff) in recv_changes.iter() {
```
utaal, line 134: I've done this locally as well to track down a potential race, and ended up just changing `messages` and `internal` to a single vector of `changes`, as this filtering can be easily done at the receiver. Any reason you prefer to keep them separate?

frankmcsherry: Oh, only that it was written that way to begin I think, and I wanted to change minimal things. Did you make the change because this was error-prone? Historically they were separate because that's how the data were presented (separate `messages` and `internal`).

frankmcsherry: I think also historically vectors of enums where kept as multiple demux'd vectors to minimize branching in the middle of hot loops (back when the time to update an operator was sub-microsecond. I think that time may have drifted up enough that this sort of attention to detail is less critical.

timely/src/progress/change_batch.rs
```diff
@@ -251,7 +272,7 @@ impl<T:Ord> ChangeBatch<T> {
     }
 
     /// Expose the internal vector of updates.
-    pub fn unstable_internal_updates(&self) -> &Vec<(T, i64)> { &self.updates }
+    pub fn unstable_internal_updates(&self) -> &[(T, i64)] { &self.updates[..] }
```
frankmcsherry, line 275: I'm not sure what this change was about, but I think this reduces functionality (we can no longer check the internal capacity). I bet I thought I was just tidying, but I think I'll skip doing this.

frankmcsherry:
Closing in favor of #352, which is just a cleaned up version of this PR.

PR #327, opened 2020-05-12
WIP: fix validate_progress in subgraph

utaal:
It's unsafe to create capabilities based on the input frontier of the operator: a negative incoming point-stamp change may be in flight, which would cause a race condition downstream.

I'm a little short on time right now, but I'd be happy to provide more details on the race condition. From preliminary testing, no operator violates this stricter requirement.

This was found as part of the verification effort at ETH, by @saradecova.


frankmcsherry:
Seems legit! When you all have more time I'd be interested in the race, and the implications to the method (this seems to now catch more error cases, but it sounds like none of the implementors were screwing this up; is this more about writing down the correct statement about progress validation than changing the system's behavior?)

PR #330, opened 2020-06-07
Fix broken link to wordcount.rs example

kachayev:


frankmcsherry:
Looks great! Thanks for the catch, and sorry about the delay.

Issue #107, opened 2017-11-08
Local validation of progress updates


frankmcsherry:
Ideally the progress API exposed by `Operate` trait, by which operators report input messages consumed, capabilities dropped and held, and output messages produced, would have some pleasant invariants about batches of updates by which we could increase our confidence in their global properties.

One intended (but not satisfied) invariant is "each newly held capability must be greater or equal to some consumed input message or held capability" and "each send output must have time greater or equal to some consumed input message or held capability".

This invariant is not currently satisfied because subgraphs eagerly report information about the future that are known to be true, but aren't justifiable yet. 

For example, consider a subgraph with two inputs, managing an operator that consumes input and may hold capabilities. Imagine this graph is executed across multiple workers. One subgraph instance can receive a report from its managed operator that it has consumed a message and now holds a capability. The subgraph currently chooses to report that information upwards (it now holds a capability) but it is not yet in a position to indicate which of the two inputs the message came in through (that information is perhaps with the other worker, who performed the ingestion and sent a progress update relating this, but it has not yet been received).

Although this information is not incorrect, nor does it lead to errors in the protocol (known errors, at least), it is nonetheless confusing from the point of view of invariants maintained. The subgraph appears to be claiming a capability without consuming any input messages. There is the intent to do so in the future, and in this case we know that the only way the report of a consumed message can arrive is through the subgraph.

Perhaps the subgraph should delay this information, even though it knows it will happen, until the progress update "makes sense". It could wait until it hears about the message that justified the capability, and only then report claiming the capability; as long as the system only needs to know about the frontier, this would be the first moment it could advance to the frontier because until this point the unacknowledged message blocked the frontier.

This would allowed us to get closer to imposing invariants on each batch of updates, which should increase confidence in the protocol, at the expense of complicating the responsibility of the subgraph. On the positive side, this buffering would result in less movement of progress updates, communicating changes less frequently.

frankmcsherry:
Recent work observes that our current progress validation was too relaxed to catch all errors (#327) but the fix ended up being too severe and caught non-errors (#331). One conclusion is that, yet again, we might want a more locally verifiable protocol.

[...]

PR #331, opened 2020-06-23
Relax pointstamp validation

frankmcsherry:
In #327 the pointstamp validation logic was made more strict, requiring either that an input message be consumed or that a capability existed. This is too strict, owing to the hierarchical nature of progress tracking.

Specifically, running differential's `bfs` the following way, with two processes
```
cargo run --example bfs -- 10 10 10 100000000 no -w1 -n2 -p0
cargo run --example bfs -- 10 10 10 100000000 no -w1 -n2 -p1
```
results in a crash.

The reason here is (I think) that a subgraph, which presents upward as an operator, may internally receive a message from another peer. The subgraph presents the existence of that message upward as an internal capability. However, the subgraph operator may not have yet received notice of any incoming message, nor hold any other capability. Nonetheless, nothing is actually wrong (other than that the protocol is hard to verify locally).

I think the summary is: prior to #327 the `validate_progress` test could pass progress information that could be invalid (as the issue notes, because races could mean that the progress information is about to change). After #327 the `validate_progress` test can fail progress information that is valid.

There is a legit complaint that the progress traffic is hard to locally verify, which I 100% agree with. I think there is a dopey answer that is "operators should buffer progress information that does not yet make sense" which is permitted because (unless there are bugs) there no safety requirements that progress traffic move at any speed.

### DIFF
--- a/kafkaesque/Cargo.toml
+++ b/kafkaesque/Cargo.toml
@@ -10,4 +10,4 @@ abomonation="0.7"
 timely = { path = "../timely" }
 
 [dependencies.rdkafka]
-version = "0.20.0"
+version = "0.23.0"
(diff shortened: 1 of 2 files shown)


frankmcsherry:
Landing this, but with an eye towards more effective validation. Issue #107 seems like a good place to conduct further discussion, and I'll record some notes there!

PR #332, opened 2020-07-03
Change naming of long-lived threads

quodlibetor:
This makes threads obviously come from timely, and is aligned with common
formatting -- no whitespace and a `:` between library name and thread purpose.

It is more common to have all threads have exactly the same name, but obviously
in most libraries threads are completely interchangeable, where they are unique
in timely.


frankmcsherry:
Seems legit! Thank you!

PR #333, opened 2020-07-06
Shave two characters off of worker thread name

benesch:
Linux limits thread names to 15 characters. This limit interacts poorly
with the naming of worker threads, as e.g. worker thread 10's name gets
truncated to "timely:worker-1".

Use "timely:work" as the prefix instead, which leaves space for three
digits of thread ID, and is nicely symmetric with the "timely:send" and
"timely:recv" threads.


frankmcsherry:
Excellent! The symmetry is obviously the best reason. :D

PR #338, opened 2020-08-30
remove an `error[E0282]: type annotations needed`

nrolland:
the T in CapabilityRef<T> can't seem to b inferred using stable or nightly rust


nrolland:
Sorry you are right, I was building up the program as it appears in the published book, where the previous excerpt has no type annotation ! 

I'll make a pull request there !

PR #342, opened 2020-11-09
add a convenience function to make named dataflows with otherwise default arguments

elindsey:
I'm naming the logging dataflows in materialize and keep writing boilerplate like

```
    let worker_logging = worker.log_register().get("timely");
    let traces = worker.dataflow_core(
        "Dataflow: differential logging",
        worker_logging,
        Box::new(()),
        move |_, scope| {});
```

Which could instead be

```
let traces = worker.dataflow_named("Dataflow: differential logging", move |scope| {});
```


frankmcsherry (approved):
Seems like a great idea, thanks!

elindsey:
When convenient, could I get a merge? 🙂

PR #343, opened 2020-11-21
Documentation improvements

petrosagg:
Did a few spelling and grammar fixes while going through the docs and also updated all links to use the new link-to-item markdown syntax.


mdbook/src/chapter_2/chapter_2_4.md
```diff
@@ -40,7 +40,7 @@ Most of what is interesting lies in the closure, so let's first tidy up some loo
...
+The input handle `input` has one primary method, `next`, which may return a pair of timestamps and a batch of data. Rust really likes you to demonstrate a commitment to only looking at valid data, and our `while` loop does what is called deconstruction: we acknowledge the optional structure and only execute in the case the `Option` variant is `Some`, containing data. The `next` method could also return `None`, indicating that there is no more data available at the moment. It is strongly recommended that you take the hint and stop trying to read inputs at that point; timely gives you the courtesy of executing whatever code you want in this closure, but if you never release control back to the system you'll break things (timely employs ["cooperative multitasking"](https://en.wikipedia.org/wiki/Cooperative_multitasking)).
```
frankmcsherry, line 43: I think is this is not correct. It is a pair of a timestamp and a batch of data. Strictly speaking, it is a pair of a *capability* (a `Capability<Timestamp>`, for linking purposes) and a batch of data. None of this is on you, though. I can clean it up in post if you prefer!

frankmcsherry:
Seems good, thanks! I left one comment that you can take/leave (I'll patch it up afterwards if you'd rather not).

petrosagg:
@frankmcsherry you're right, I missed that. I just pushed a fix for that

frankmcsherry:
Thanks very much (and thanks for checking the actual type; oops!).

PR #350, opened 2021-01-11
Provide for worker-configuration parameters

benesch:
@frankmcsherry, as discussed on Slack! No promises that any of this is appealing!

Tweak the execution API so that worker configuration parameters can be
specified along with communication fabric configuration parameters. In
particular, this permits the DEFAULT_PROGRESS_MODE environment variable
to be replaced with a typed ProgressMode enum, and for its value to be
controllable with a command-line option.

The new WorkerConfig struct also admits arbitrary typed configuration
parameters, which downstream libraries (e.g., differential) can use to
store their own worker configuration variables.

### DIFF
--- a/communication/src/initialize.rs
+++ b/communication/src/initialize.rs
@@ -18,7 +18,7 @@ use logging_core::Logger;
 
 
 /// Possible configurations for the communication infrastructure.
-pub enum CommunicationConfig {
+pub enum Config {
     /// Use one thread.
     Thread,
     /// Use one process with an indicated number of threads.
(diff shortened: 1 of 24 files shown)


communication/src/initialize.rs
```diff
@@ -18,7 +18,7 @@ use logging_core::Logger;
 
 
 /// Possible configurations for the communication infrastructure.
-pub enum Configuration {
+pub enum CommunicationConfig {
```
frankmcsherry, line 21: I feel like the "recommended style" is leaving this as `Configuration` but then importing it as `CommunicationConfig` in `timely`. I don't think it matters too much outside of this crate (as it will be `CommunicationConfig` by timely users). Any thoughts?

timely/src/worker.rs
```diff
@@ -16,11 +18,79 @@ use crate::progress::operate::Operate;
...
+    }
+}
+
+/// Worker configuration.
+#[derive(Debug, Default, Clone)]
+pub struct WorkerConfig {
+    /// The progress mode to use.
+    pub(crate) progress_mode: ProgressMode,
+    /// A map from parameter name to typed parameter values.
+    registry: HashMap<String, Arc<dyn Any + Send + Sync>>,
```
frankmcsherry, line 54: Is there a non-obviously reason that this is an `Arc` instead of a `Box`?

frankmcsherry: Alternately (if it is like, setting up the configs for everyone and I've missed that) is the `Send` required for `Arc`?

timely/src/execute.rs
```diff
@@ -279,23 +369,29 @@ pub fn execute_from_args<I, T, F>(iter: I, func: F) -> Result<WorkerGuards<T>,St
...
+    worker_config: WorkerConfig,
+    func: F,
+) -> Result<WorkerGuards<T>, String>
 where
     A: AllocateBuilder+'static,
     T: Send+'static,
     F: Fn(&mut Worker<<A as AllocateBuilder>::Allocator>)->T+Send+Sync+'static {
     initialize_from(builders, others, move |allocator| {
-        let mut worker = Worker::new(allocator);
+        let mut worker = Worker::new(worker_config.clone(), allocator);
```
frankmcsherry, line 394: This looks like the `Arc` moment!

timely/src/execute.rs
```diff
@@ -269,8 +302,65 @@ pub fn execute_from_args<I, T, F>(iter: I, func: F) -> Result<WorkerGuards<T>,St
     where I: Iterator<Item=String>,
           T:Send+'static,
           F: Fn(&mut Worker<Allocator>)->T+Send+Sync+'static, {
-    let configuration = Configuration::from_args(iter)?;
-    execute(configuration, func)
+    let mut opts = getopts::Options::new();
```
frankmcsherry, line 305: The opts parsing .. I was sort of imagining as part of `ExecuteConfig`. That way e.g. someone could parse up some arguments, and then maybe flip a few, read out some and print them on screen, stuff like that. Is there a reason to stash this logic here, vs in some constructor/initialization for `ExecuteConfig`?

timely/src/worker.rs
benesch: Yeah, what we actually need is a clonable `Box<dyn Any>`, but I don't know how to make that happen easily types besides `Arc<dyn Any + Send + Sync>`, which is a bit overspecific. Specifically, I don't think `Sync` is important, but `Arc<T>` doesn't implement `Send` unless the underlying type implements `Sync` too.

I'm sure you could make something work with https://docs.rs/dyn-clone/1.0.4/dyn_clone/, but it's always a big pain in the butt in my experience.

timely/src/execute.rs
benesch: Yep!

benesch:
I overhauled the naming of these structs so that everything is called `Config` where it is defined, and gets a more specific prefix only when re-exported at the root of the `timely` crate. I also reworked the option parsing so that it's possible to parse _just_ the `timely::CommunicationConfig` from arguments or _just_ the `timely::WorkerConfig` from arguments. This requires some cooperation to make sure that none of the command-line options conflict, but I think that's ok.

frankmcsherry:
Thanks very much! I think this all looks great!

benesch:
Woohoo! I think all that's missing then is documentation of what the different progress modes actually do.

frankmcsherry:
I'm happy to add that in post, if that works for you.

benesch:
It most certainly does!

frankmcsherry:
https://github.com/TimelyDataflow/timely-dataflow/commit/bafcd7db13b0530c75e69cbc47437f59c9a0ed50

timely/src/execute.rs
```diff
@@ -181,11 +255,11 @@ where
         });
     }
 
-    let (allocators, other) = config.try_build()?;
+    let (allocators, other) = config.communication.try_build()?;
 
     initialize_from(allocators, other, move |allocator| {
 
-        let mut worker = Worker::new(allocator);
+        let mut worker = Worker::new(WorkerConfig::default(), allocator);
```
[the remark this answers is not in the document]
frankmcsherry: Probably! Lemme look further. Sorry, this stuff is definitely under-exercised.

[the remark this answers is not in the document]
frankmcsherry: Yes, it should be fixed now, with a test in place. Thanks very much for noticing!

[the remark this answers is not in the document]
sdht0: Great, cheers!

PR #351, opened 2021-01-11
Change default progress mode to demand

benesch:
The docs recommend demand mode, so it seemed like that should
be the default! Also correct some minor typos in the docs.


frankmcsherry:
What an embarrassing pile of typos!

frankmcsherry:
This looks good, and for the record the flip of the default progress mode seems good and lines up with the intended plans going forward (in case anyone ends up here because it worked out less well for them).

PR #353, opened 2021-01-19
Separate the progress logging stream, use dyn trait instead of String for timestamps

utaal:
This only supports cfg(not(feature = "bincode")), bincode support upcoming.
Currently based on the `log_progress_accurately` branch.

@frankmcsherry

### DIFF
--- a/timely/src/logging.rs
+++ b/timely/src/logging.rs
@@ -107,6 +107,11 @@ impl<T: crate::Data + std::fmt::Debug + std::any::Any> ProgressEventTimestamp fo
 }
 
 /// A vector of progress updates in logs
+///
+/// This exists to support upcasting of the concrecte progress update vectors to
+/// `dyn ProgressEventTimestamp`. Doing so at the vector granularity allows us to
+/// use a single allocation for the entire vector (as opposed to a `Box` allocation
+/// for each dynamically typed element).
 pub trait ProgressEventTimestampVec: std::fmt::Debug + std::any::Any {
     /// Iterate over the contents of the vector
     fn iter<'a>(&'a self) -> Box<dyn Iterator<Item=(&'a usize, &'a usize, &'a dyn ProgressEventTimestamp, &'a i64)>+'a>;
--- a/timely/src/logging.rs
+++ b/timely/src/logging.rs
@@ -77,6 +77,16 @@ pub trait ProgressEventTimestamp: std::fmt::Debug + std::any::Any {
     /// Upcasts this `ProgressEventTimestamp` to `Any`.
     ///
     /// NOTE: This is required until https://github.com/rust-lang/rfcs/issues/2765 is fixed
+    ///
+    /// # Example
+    /// ```rust
+    /// let ts = vec![(0usize, 0usize, (23u64, 10u64), -4i64), (0usize, 0usize, (23u64, 11u64), 1i64)];
+    /// let ts: &timely::logging::ProgressEventTimestampVec = &ts;
+    /// for (n, p, t, d) in ts.iter() {
+    ///     print!("{:?}, ", (n, p, t.as_any().downcast_ref::<(u64, u64)>(), d));
+    /// }
+    /// println!();
+    /// ```
     fn as_any(&self) -> &dyn std::any::Any;
 
     /// Returns the name of the concrete type of this object.
--- a/timely/src/logging.rs
+++ b/timely/src/logging.rs
@@ -6,6 +6,8 @@ pub type WorkerIdentifier = usize;
 pub type Logger<Event> = crate::logging_core::Logger<Event, WorkerIdentifier>;
 /// Logger for timely dataflow system events.
 pub type TimelyLogger = Logger<TimelyEvent>;
+/// Logger for timely dataflow progress events (the "timely/progress" log stream).
+pub type TimelyProgressLogger = Logger<TimelyProgressEvent>;
 
 use std::time::Duration;
 use crate::dataflow::operators::capture::{Event, EventPusher};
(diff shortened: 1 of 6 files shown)


timely/src/logging.rs
```diff
@@ -70,9 +72,44 @@ pub struct ChannelsEvent {
     pub target: (usize, usize),
 }
 
-#[derive(Serialize, Deserialize, Abomonation, Debug, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
+#[cfg(not(feature = "bincode"))]
+/// Encapsulates (de)serialization, Debug, Display, for dynamically typed timestamps in logs
+pub trait ProgressEventTimestamp: std::fmt::Debug + std::fmt::Display + std::any::Any {
+    /// Encodes a typed reference into a binary buffer.
```
frankmcsherry, line 78: Just to check, `Timestamp` has already `Debug+Any+Abomonation/Serialize`. Is the motivation for the trait to add `Display`, or because the `communication::Data` trait isn't actually helpful?

frankmcsherry:
Thanks very much!

I think this looks good, though I have to admit not knowing about ease/pain of logging out. Strings look pretty good, but other formats could be hard and I wouldn't know. This does the good things (imo) of extracting the progress logging to a new stream so that it can be turned off cheaply. I had a comment or two about the rationale for some traits, not because I doubt the rationale but more to write it down so that in a few months I can remember why it wasn't another way. :)

timely/src/logging.rs
utaal: It’s because Data can’t be turned into a trait, because entomb takes a type parameter. Should probably add a comment to remind future readers.

utaal:
I changed things to be serialization-agnostic, but to still offer the option of recovering the real timestamp type in the registered log function. This way, someone who needs detailed, typed information, can downcast to the concrete types of the timestamp present in the dataflow.

I also actually sent native types and not formatted strings, which I had failed to do previously (in `Progcaster`).

timely/src/logging.rs
```diff
@@ -70,9 +72,58 @@ pub struct ChannelsEvent {
     pub target: (usize, usize),
 }
 
-#[derive(Serialize, Deserialize, Abomonation, Debug, Clone, Hash, Eq, PartialEq, Ord, PartialOrd)]
+/// Encapsulates Any and Debug for dynamically typed timestamps in logs
+pub trait ProgressEventTimestamp: std::fmt::Debug + std::any::Any {
```
frankmcsherry, line 76: Random comment, not a change but: I bet this trait eventually becomes useful beyond progress events. At least, assuming there is any other reason to report information about timestamps (e.g., maybe other timestamp things, I dunno) they'll probably make use of this too. Again: nothing to change, just thinking out loud.

timely/src/logging.rs
```diff
@@ -70,9 +72,58 @@ pub struct ChannelsEvent {
...
+    /// `"std::option::Option<std::string::String>"`.
+    fn type_name(&self) -> &'static str;
+}
+impl<T: crate::Data + std::fmt::Debug + std::any::Any> ProgressEventTimestamp for T {
+    fn as_any(&self) -> &dyn std::any::Any { self }
+
+    fn type_name(&self) -> &'static str { std::any::type_name::<T>() }
+}
+
+/// A vector of progress updates in logs
```
frankmcsherry, line 109: Minor, but this would be a great place to explain the need for a trait (vs a `type` or a `struct` that matches the same description).

timely/src/progress/broadcast.rs
```diff
@@ -58,34 +59,26 @@ impl<T:Timestamp+Send> Progcaster<T> {
...
-            self.logging.as_ref().map(|l| {
+            self.progress_logging.as_ref().map(|l| {
 
                 // Pre-allocate enough space; we transfer ownership, so there is not
                 // an apportunity to re-use allocations (w/o changing the logging
                 // interface to accept references).
-                let mut messages = Vec::with_capacity(changes.len());
-                let mut internal = Vec::with_capacity(changes.len());
+                let mut messages = Box::new(Vec::with_capacity(changes.len()));
+                let mut internal = Box::new(Vec::with_capacity(changes.len()));
```
frankmcsherry, line 68: I'm super confused that this works. I would have thought that as soon as this is boxed up you lose access to the `push` methods. Or is it because there is an auto-conversion from `Box<T>` to `Box<dyn Trait>` for any `T: Trait`? No change needed, just surprising to me!

timely/src/progress/broadcast.rs
```diff
@@ -138,32 +131,31 @@ impl<T:Timestamp+Send> Progcaster<T> {
...
+                l.log(crate::logging::TimelyProgressEvent {
                     is_send: false,
                     source: source,
                     seq_no: counter,
                     channel,
                     addr: addr.clone(),
-                    messages,
-                    internal,
+                    messages: Box::new(messages),
+                    internal: Box::new(internal),
```
frankmcsherry, line 158: Yeah, this is what I thought you would need to do. Is there a reason this is needed here but not previously? If so I think we should doc it, and if not we should unify up the two cases to look the same (maybe just aesthetics, but also to avoid randoms (myself) asking this question in the future).

frankmcsherry (approved):
I think this looks good. I have some questions, but none of them need to block merging this in to the PR (and potentially then merging the PR). I'll not merge right away in case any of the comments seem actionable to you, but otherwise I'm happy to do it now!

timely/src/logging.rs
utaal: Yea, I thought the same when I wrote it, but it seemed unnecessary to push it to a separate module at this stage.

timely/src/progress/broadcast.rs
utaal: The asymmetry was unintentional (and I guess a potential perf defect). Fixed in the last commit.

timely/src/progress/broadcast.rs
utaal: I believe `push` works because of `https://doc.rust-lang.org/std/boxed/struct.Box.html#impl-Deref` (`Deref` impl for `Box`). And yes, you can upcast a `Box<T>` to a `Box<dyn Trait>` for the same reason you can upcast a `&T` to a `&dyn Trait` for `T: Trait`.

utaal: To clarify, the type of `messages` and `internal` is still the concrete, non-`dyn` vector type. The upcast happens on assignment to the target struct below.

timely/src/logging.rs
utaal: Added to the doc in the latest commit.

utaal:
I believe I addressed the comments: so go ahead and merge if the latest changes look good! Thanks for the review.

PR #354, opened 2021-01-25
Fix the link to timely-dataflow documentation

Cecca:
Point it to the latest version


frankmcsherry:
That makes sense; thanks very much!

PR #357, opened 2021-02-07
Add support for native async stream sources

petrosagg:
Example usage:

```rust
use timely::dataflow::operators::{Event, ToStreamAsync};
use timely::dataflow::operators::Inspect;

#[tokio::main]
async fn main() {
    let iter = (0..4)
        .map(|x| Event::Message(0, x))
        .chain(Some(Event::Progress(1)));
    let stream = futures::stream::iter(iter);

    timely::example(move |scope| {
        let (future, stream) = stream.to_stream(scope);
        tokio::spawn(future);
        stream.inspect(|x| println!("seen: {:?}", x));
    });
}
```

### DIFF
--- a/timely/src/dataflow/operators/capability.rs
+++ b/timely/src/dataflow/operators/capability.rs
@@ -371,10 +371,14 @@ impl<T: Timestamp> CapabilitySet<T> {
     /// Downgrades the set of capabilities to correspond with the times in `frontier`.
     ///
     /// This method panics if any element of `frontier` is not greater or equal to some element of `self.elements`.
-    pub fn downgrade(&mut self, frontier: &[T]) {
+    pub fn downgrade<B, F>(&mut self, frontier: F)
+    where
+        B: std::borrow::Borrow<T>,
+        F: IntoIterator<Item=B>,
+    {
         let count = self.elements.len();
-        for time in frontier.iter() {
-            let capability = self.delayed(time);
+        for time in frontier.into_iter() {
+            let capability = self.delayed(time.borrow());
             self.elements.push(capability);
         }
         self.elements.drain(..count);
--- a/timely/src/dataflow/operators/to_stream.rs
+++ b/timely/src/dataflow/operators/to_stream.rs
@@ -1,15 +1,14 @@
 //! Conversion to the `Stream` type from iterators.
 
-use std::future::Future;
 use std::pin::Pin;
-use std::sync::{Arc, Mutex};
-use std::task::{Context, Poll, Waker};
+use std::sync::Arc;
+use std::task::{Context, Poll};
 
 use crate::dataflow::channels::Message;
 use crate::dataflow::operators::generic::operator::source;
+use crate::dataflow::operators::CapabilitySet;
 use crate::dataflow::{Scope, Stream};
 use crate::progress::Timestamp;
-use crate::scheduling::SyncActivator;
 use crate::Data;
 
 /// Converts to a timely `Stream`.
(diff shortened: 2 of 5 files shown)


frankmcsherry:
Thanks! Can you say a bit about `futures-util` and how important it is? It looks like it pulls in a fair bit non-optionally.

timely/src/dataflow/operators/to_stream.rs
```diff
@@ -56,3 +63,215 @@ impl<T: Timestamp, I: IntoIterator+'static> ToStream<T, I::Item> for I where I::
...
+    SyncActivationError,
+}
+
+/// Data and progress events of the native stream.
+pub enum Event<T, D> {
+    /// Indicates that timestamps have advanced to time T
+    Progress(T),
+    /// Indicates that event D happened at time T
+    Message(T, D),
+}
```
frankmcsherry, line 83: We should talk through this! There is already a similar enum in the `capture` module, and it is a bit more rich than this. In particular, this enum isn't able to capture partially ordered times, where there may be a set of capabilities rather than just one.

petrosagg:
> Thanks! Can you say a bit about `futures-util` and how important it is? It looks like it pulls in a fair bit non-optionally.

It's still smaller than the full `futures` crate but I just pushed a change that reduces it to `futures-core`. The only reason I needed `futures-util` is for the `.fuse()` method to match the implementation on iterators but turns out I don't need that

timely/src/dataflow/operators/to_stream.rs
petrosagg: I did see that one but wasn't sure what the Vec is for. Let's talk about it tomorrow

frankmcsherry:
Thanks! I'll get cracking on this promptly. The main ? for me is getting my head around `Future` so I can own some of that technical debt. I'm familiar with the C# version of it, but I'll need to get up to speed on Rust's version.

timely/src/dataflow/operators/capability.rs
```diff
@@ -371,10 +371,14 @@ impl<T: Timestamp> CapabilitySet<T> {
     /// Downgrades the set of capabilities to correspond with the times in `frontier`.
     ///
     /// This method panics if any element of `frontier` is not greater or equal to some element of `self.elements`.
-    pub fn downgrade(&mut self, frontier: &[T]) {
+    pub fn downgrade<B, F>(&mut self, frontier: F)
+    where
+        B: std::borrow::Borrow<T>,
+        F: IntoIterator<Item=B>,
```
frankmcsherry, line 377: Minor, but can this be
```rust
F: IntoIterator,
F::Item: std::borrow::Borrow<T>,
```
or is the generic argument `B` necessary / helpful?

timely/src/dataflow/operators/to_stream.rs
```diff
@@ -56,3 +60,108 @@ impl<T: Timestamp, I: IntoIterator+'static> ToStream<T, I::Item> for I where I::
...
+pub enum Event<F: IntoIterator, D> {
+    /// Indicates that timestamps have advanced to frontier F
+    Progress(F),
+    /// Indicates that event D happened at time T
+    Message(F::Item, D),
+}
+
+/// Converts to a timely `Stream`.
+pub trait ToStreamAsync<T: Timestamp, D: Data> {
+    /// Converts a [native `Stream`](futures_util::stream::Stream) of [`Event`s](Event) into a [timely
```
frankmcsherry, line 74: Now that this doesn't return a `TimelyFuture`, should it just be an implementation of `ToStream` for `Pin<Box<Stream>>`?

petrosagg: I believe this will hit the specialisation issues since `ToStream` is already implemented for a generic `I`

timely/src/dataflow/operators/to_stream.rs
```diff
@@ -56,3 +60,108 @@ impl<T: Timestamp, I: IntoIterator+'static> ToStream<T, I::Item> for I where I::
...
+                        Some(Event::Progress(time)) => {
+                            cap_set.downgrade(time);
+                        }
+                        Some(Event::Message(time, data)) => {
+                            output.session(&cap_set.delayed(&time)).give(data);
+                        }
+                        None => {
+                            cap_set.downgrade(&[]);
+                            break;
+                        }
```
frankmcsherry, line 129: This is interesting! I suppose it is a reasonable behavior to error-proof use of the stream. It duplicates the information that `Event::Progress([])` represents, but I can't think what else it would mean to end a stream. Elsewhere `None` just means "nothing else at the moment", but since you can presume that the stream won't ever pause like that, I guess we get more information to return.

timely/src/dataflow/operators/to_stream.rs
frankmcsherry: Ah, and `Stream` implements `Iterator`? I can't actually tell for its docs.

petrosagg:
ok, I did some minor cleanups and fixed the example code. It should be good to merge I think!

timely/src/dataflow/operators/to_stream.rs
benesch: `Stream` doesn't implement `Iterator`, that's the problem! But because Rust you can't have both of:

```
impl<I> ToStream for I where I: Iterator {}
impl<I> ToStream for I where I: Stream {}
```

frankmcsherry: Oh sorry I thought `Stream` was a struct for some reason. Nvm!

benesch (approved):
The futures goop looks great to me. No opinion on the `Event` protocol!

timely/src/dataflow/operators/to_stream.rs
```diff
@@ -56,3 +60,82 @@ impl<T: Timestamp, I: IntoIterator+'static> ToStream<T, I::Item> for I where I::
...
+                let mut context = Context::from_waker(&waker);
+
+                // Consume all the ready items of the source_stream and issue them to the operator
+                while let Poll::Ready(item) = self.as_mut().poll_next(&mut context) {
+                    match item {
+                        Some(Event::Progress(time)) => {
+                            cap_set.downgrade(time);
+                        }
+                        Some(Event::Message(time, data)) => {
+                            output.session(&cap_set.delayed(&time)).give(data);
```
frankmcsherry, line 130: Minor, but this will eventually have throughput issues. At least, timely can send data around much faster than it can mint capabilities. It may eventually be smart to have the `D` of the enum implement `IntoIterator<Item = D2>`, though perhaps .. that is just a flatmap after this. Idk. Doesn't block anything here, but wanted to call out!

frankmcsherry (approved):
Looks good to me. Various comments that are good for future discussion (lol not intended nvm sry).

PR #359, opened 2021-02-17
Accept subgraph frontiers in setup

frankmcsherry:
This PR fixes a bug that mysteriously hasn't shown up until some stress testing over in https://github.com/MaterializeInc/materialize/issues/5622.

The issue is that in set-up, a `Subgraph` will propagate pointstamp information among its operators and then schedule them for their first time. However, it would not first accept updates from the outside world, which meant that the initial calls into operator logic would not reflect "safe" frontiers, especially if an operator's input was from outside the scope (there would be no initial capability locally within the subgraph to present at it). The observed behavior was a spurious empty frontier in the first call, followed by legit frontiers after that call.

In fact, this manifests in `enterleave::enter_at` with enough panicky double-checking code (and indeed that example has an operator that consumes directly from outside the scope).

This fix is very local, but another take is that `set_external_summary` could plausibly be deleted, as the surrounding scope no longer presents a summary downwards, and initial frontiers are presented in the initial population of the `SharedProgress` struct rather than by method call.

### DIFF
--- a/timely/examples/logging-send.rs
+++ b/timely/examples/logging-send.rs
@@ -5,7 +5,7 @@ use timely::dataflow::operators::{Input, Exchange, Probe};
 
 // use timely::dataflow::operators::capture::EventWriter;
 // use timely::dataflow::ScopeParent;
-use timely::logging::TimelyEvent;
+use timely::logging::{TimelyEvent, TimelyProgressEvent};
 
 fn main() {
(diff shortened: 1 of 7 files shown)


frankmcsherry:
Hrm. Mysteriously I have picked up some branch I must have been on. Closing and re-opening with the one line diff. :)

PR #370, opened 2021-03-08
Pre-allocate vectors

Kixiron:
Pre-allocates more vectors in the hopes of reducing startup times and memory usage

### DIFF
--- a/communication/src/allocator/process.rs
+++ b/communication/src/allocator/process.rs
@@ -38,7 +38,7 @@ impl AllocateBuilder for ProcessBuilder {
             let buzzer = Buzzer::new();
             worker.send(buzzer).expect("Failed to send buzzer");
         }
-        let mut buzzers = Vec::new();
+        let mut buzzers = Vec::with_capacity(self.buzzers_recv.len());
         for worker in self.buzzers_recv.iter() {
             buzzers.push(worker.recv().expect("Failed to recv buzzer"));
         }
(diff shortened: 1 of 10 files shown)


communication/src/allocator/process.rs
```diff
@@ -69,19 +69,19 @@ pub struct Process {
...
+    pub fn inner(&mut self) -> &mut Thread { &mut self.inner }
     /// Allocate a list of connected intra-process allocators.
     pub fn new_vector(peers: usize) -> Vec<ProcessBuilder> {
 
-        let mut counters_send = Vec::new();
-        let mut counters_recv = Vec::new();
-        for _ in 0 .. peers {
+        let mut counters_send = Vec::with_capacity(peers);
+        let mut counters_recv = Vec::with_capacity(peers);
+        for _ in 0..peers {
```
frankmcsherry, line 78: Ahhh. Was looking so good until here. :D

timely/src/dataflow/operators/partition.rs
```diff
@@ -25,16 +25,17 @@ pub trait Partition<G: Scope, D: Data, D2: Data, F: Fn(D)->(u64, D2)> {
     fn partition(&self, parts: u64, route: F) -> Vec<Stream<G, D2>>;
 }
 
-impl<G: Scope, D: Data, D2: Data, F: Fn(D)->(u64, D2)+'static> Partition<G, D, D2, F> for Stream<G, D> {
+impl<G: Scope, D: Data, D2: Data, F: Fn(D) -> (u64, D2) + 'static> Partition<G, D, D2, F>
+    for Stream<G, D>
+{
```
frankmcsherry, line 30: I think `cargo fmt` line noise?

frankmcsherry:
Thanks! Generally looks good, in particular the `ref` and `ref mut` removal, and the `with_capacity` calls which much more clearly indicate intent. There is some `cargo fmt` line noise that it would be great to ditch (the code is formatted for a wide and short display, and `cargo fmt`'s opinion that I shouldn't have that display is ...).

frankmcsherry:
I'm up for looking in to what is the source of the slowness. I've profiled it for some microbenchmarks, and .. it was smeared out a bit. There are several places where allocation could be more lazy (e.g. data buffers between workers) and where fewer allocations could be used (e.g. in reachability). But, if you end up with some traces that incriminate things, I'm up for talking through what is needed and what is not.

PR #371, opened 2021-03-09
Correctly propagate worker config

frankmcsherry:
This looks like an oversight that was never tested.

cc: @sdht0


frankmcsherry:
I added two tests (well, the same test twice) that confirm that the configuration was not being propagated before, but is being propagated now.

PR #373, opened 2021-03-20
Fix warning: unnecessary parentheses around type

ihrwein:


ihrwein:
Sorry, this should have been against the old repo. I was following the mdbook and came across this warning: https://timelydataflow.github.io/timely-dataflow/chapter_0/chapter_0_1.html

PR #375, opened 2021-03-30
Reachability logging

frankmcsherry:
This PR is an attempt at replicating #321 with some firm isolation between existing logging machinery and any new logging machinery. It has some opinions about where and how frequently we should log things (in batch, after consolidation), and it uses the abstraction of #352 to allow timestamps to be recovered via `dyn Any` downcasting.

cc @utaal for any thoughts, but I should be able to move this forward even without any particular comments. I think if I extend it to capture the topology as well as the events it pretty closely resembles the intent of #321.

### DIFF
--- a/timely/src/progress/reachability.rs
+++ b/timely/src/progress/reachability.rs
@@ -27,7 +27,7 @@
 //! builder.add_edge(Source::new(2, 0), Target::new(0, 0));
 //!
 //! // Construct a reachability tracker.
-//! let (mut tracker, _) = builder.build();
+//! let (mut tracker, _) = builder.build(None);
 //!
 //! // Introduce a pointstamp at the output of the first node.
 //! tracker.update_source(Source::new(0, 0), 17, 1);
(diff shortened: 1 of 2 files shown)


utaal:
This looks great! Glad the `dyn Any` abstraction is working out here too.

frankmcsherry:
Sweet. I think I'll land it. The topology information is a bit harder to capture in that `T::Summary` isn't as trivially packed into the timestamp vector trait. I think it could be put behind the same sort of abstraction (perhaps the same one, if we removed "timestamp" from its name) but we'll need to add that in the future (no worries to do so, and you can grab the edge information out of vanilla timely logging, minus any `T::Summary` decorations).

frankmcsherry:
> Topology capture was, I believe, essentially just Summaryies and timestamp type information: do you think you'd add those to the other new progress logging events or here directly?

Probably the right thing to do is log them here. Like a `CreateTopology(...)` event that signifies the start, and perhaps a `ShutDown` event at the end so that folks can clean up. I stalled a bit on figuring out the right way to put a new trait in place for `dyn` stuff, and wanted to try it out to see how well what was here worked for some use cases.

PR #376, opened 2021-03-31
Some small nits

victorzz0:
I saw a couple of small nits that can be fixed easily, but these are the only ones that I remember now...


frankmcsherry:
These look good, thank you!

PR #379, opened 2021-04-10
Usability and performance tweaks

Kixiron:
* Added some `Default` implementations to types where possible
* Removed some redundant clones
* Addressed a few minor lints
* Swapped from indexing to iteration where possible
* Added methods to `Antichain` & `ChangeBatch` that allow pre-allocating space within them
* Made `summarize_outputs()` do as much pre-allocation as possible instead of just repeatedly pushing, should help to cut down the number of alloc calls it makes
* Moved trait bounds on `Antichain`, `MutableAntichain` & `ChangeBatch` from the type declaration to the methods that need it, this is really nice from a user perspective because it means you can use these types within your own structs without having to add trait bounds *everywhere*
* Made `MutableAntichain::update_iter()` use [`Iterator::size_hint()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint) to pre-allocate space for the new elements


frankmcsherry (approved):
These all look good thanks very much!

PR #380, opened 2021-04-10
Added the FilterMap operator

Kixiron:
Adds the monadic `FilterMap` operator which allows filtering & mapping a stream.
The way the trait is designed (using the associated type `Output` for the method's return value) allows the operator to be generally applicable, meaning that this same `FilterMap` operator can be implemented for `differential_dataflow::Collection`

```rust
let filtered = ["10", "11", "12", "invalid number"]
    .to_stream(scope)
    .filter_map(|x| x.parse::<i32>().ok());
```

### DIFF
--- a/timely/src/dataflow/operators/filter_map.rs
+++ b/timely/src/dataflow/operators/filter_map.rs
@@ -1,0 +1,96 @@
+//! Filters and maps a [`Stream`] using the given predicate
+
+use crate::{
+    dataflow::{channels::pact::Pipeline, operators::Operator, Scope, Stream},
+    Data,
+};
+
+/// Filters and maps a [`Stream`] at the same time
+pub trait FilterMap<D, D2> {
+    /// The output type of the operator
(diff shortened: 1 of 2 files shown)


frankmcsherry:
Thanks for the PR!

Do you have thoughts about the motivation here? My initial disposition is that this is covered by `flat_map`, and increasing the testing/compiling/maintaining surface area is net-negative without a good reason. I'm up for being convinced, but if it all just the same as `flat_map` other than the name, I'd rather not. Any strong feelings about that?

Kixiron:
It's mainly a user experience thing, having a dedicated `.filter_map()` operator matches the apis of iterators and streams which rust users are (presumably) already familiar with. I know that I personally took (too long) to realize that `flat_map` was the replacement since I just expected a `filter_map` operator to exist just like iterators and such. I've also ended up re-implementing it in every project I've made, so I figured it was worth inclusion

PR #381, opened 2021-04-12
Fix description of iteration example

aljoscha:
I was also wondering why we're using `feedback` instead of `loop_variable` here. But this at least changes the text to match the example.

### DIFF
--- a/mdbook/src/chapter_4/chapter_4_2.md
+++ b/mdbook/src/chapter_4/chapter_4_2.md
@@ -34,7 +34,7 @@ fn main() {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


aljoscha:
Also added a fix for another example description.

frankmcsherry:
Sorry for the hold up here; I dropped the ball!

The text changes look good. There has been historical "confusion" between `feedback` and `loop_variable` (at one point names shifted to match functionality) and I suspect that confused leaked through some explanatory text.

aljoscha:
No worries! 😅

PR #383, opened 2021-04-14
Fixed bug in Subgraph::get_internal_summary()

Kixiron:
Fixes a bug introduced in #379 that caused too few elements/sub-elements being added to the internal summary

### DIFF
--- a/timely/src/progress/subgraph.rs
+++ b/timely/src/progress/subgraph.rs
@@ -637,8 +637,17 @@ impl<T: Timestamp> PerOperatorState<T> {
 
         let (internal_summary, shared_progress) = scope.get_internal_summary();
 
-        assert_eq!(internal_summary.len(), inputs);
-        assert!(!internal_summary.iter().any(|x| x.len() != outputs));
+        assert_eq!(
+            internal_summary.len(),
+            inputs,
+            "operator summary has {} inputs when {} were expected",
+            internal_summary.len(),
+            inputs,
+        );
+        assert!(
+            !internal_summary.iter().any(|x| x.len() != outputs),
+            "operator summary had too few outputs",
+        );
 
         PerOperatorState {
             name:               scope.name().to_owned(),
--- a/timely/src/progress/subgraph.rs
+++ b/timely/src/progress/subgraph.rs
@@ -530,24 +530,22 @@ where
         assert_eq!(self.children[0].outputs, self.inputs());
         assert_eq!(self.children[0].inputs, self.outputs());
 
-        let mut internal_summary = Vec::with_capacity(self.inputs());
-        for summary in self.scope_summary.iter() {
-            let scope_summary = summary.iter()
-                .map(|output| {
-                    output.elements()
-                        .iter()
-                        .cloned()
-                        .fold(
-                            Antichain::with_capacity(output.elements().len()),
-                            |mut antichain, path_summary| {
-                                antichain.insert(TInner::summarize(path_summary));
-                                antichain
-                            },
-                        )
-                })
-                .collect();
+        let mut internal_summary: Vec<Vec<_>> = (0..self.inputs())
+            .map(|_| (0..self.outputs()).map(|_| Antichain::new()).collect())
+            .collect();
 
-            internal_summary.push(scope_summary);
+        for (summary, internal_summary) in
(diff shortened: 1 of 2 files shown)


timely/src/progress/subgraph.rs
```diff
@@ -530,26 +530,37 @@ where
...
-                })
-                .collect();
-
-            internal_summary.push(scope_summary);
+        // Note that we need to have `self.inputs()` elements in the summary
+        // with each element containing `self.outputs()` antichains regardless
+        // of how long `self.scope_summary` is
+        let mut internal_summary: Vec<Vec<_>> = (0..self.inputs())
+            .map(|_| (0..self.outputs()).map(|_| Antichain::new()).collect())
+            .collect();
```
frankmcsherry, line 538: Do you prefer this to 
```rust
let mut internal_summary = vec![vec![Antichain::new(); self.outputs()]; self.inputs()];
```
or was it just a matter that it was easier to type the above vs recover the prior code?

timely/src/progress/subgraph.rs
```diff
@@ -530,26 +530,37 @@ where
...
+            internal_summary.reserve(summary.len());
+
+            for (output, antichain) in summary.iter().zip(internal_summary) {
+                antichain.reserve(output.elements().len());
+
+                for path_summary in output.elements() {
+                    antichain.insert(TInner::summarize(path_summary.clone()));
+                }
+            }
         }
```
frankmcsherry, line 552: This might be because of the recent confusion, but at the moment I def prefer something more explicit like what was here before
```
        for input in 0 .. self.scope_summary.len() {
            for output in 0 .. self.scope_summary[input].len() {
                for path_summary in self.scope_summary[input][output].elements().iter() {
                    internal_summary[input][output].insert(TInner::summarize(path_summary.clone()));
                }
            }
        }
```
vs the zipping of iterators that might terminate early if they don't have the same lengths, etc.

timely/src/progress/subgraph.rs
```diff
@@ -530,26 +530,37 @@ where
...
 
+        debug_assert_eq!(
+            internal_summary.len(),
+            self.inputs(),
+            "the internal summary should have as many elements as there are inputs",
+        );
+        debug_assert!(
+            internal_summary.iter().all(|summary| summary.len() == self.outputs()),
+            "each element of the internal summary should have as many elements as there are outputs",
+        );
```
frankmcsherry, line 562: These are good, but tbh we should probably steal them for the assert that was failing, so that they apply to all operators rather than just subgraphs.

timely/src/progress/subgraph.rs
Kixiron: The above pre-allocates (and initializes) the proper number of elements so that no further reallocations have to be made

timely/src/progress/subgraph.rs
Kixiron: Well, the thing is that iteration (without indices) is faster than indexing because of the whole bounds checking thing, checks are never emitted with iteration/zipping but with manual indexing sometimes llvm drops the ball, it's your call though

timely/src/progress/subgraph.rs
Kixiron: I'm not sure what you mean by that, you mean to move the assertions to another place?

timely/src/progress/subgraph.rs
frankmcsherry: I think this does too. The `vec![thing; len]` desugars down to roughly the same code with a `thing.clone()` instead of `Antichain::new()`.

timely/src/progress/subgraph.rs
frankmcsherry: This is fine, sorry. I think it can be statically verified at this point (with the explicit builder). I was thinking that `PerOperatorState::new()` is the place that one might want to re-verify, but there is already the assert there (without the nice error statements).

timely/src/progress/subgraph.rs
frankmcsherry: Let's go with clarity for the moment, and if we end up seeing CPU being a problem here ever, we could revisit it.

frankmcsherry:
This looks great; thanks very much!

PR #384, opened 2021-04-19
Worked on Capability egronomics

Kixiron:
* Removed the `mint()` and `mint_ref()` methods in favor of `pub(crate)` `new()` constructors
* Implemented `Debug` for `ActivateCapability` & `Activations`
* Added `.try_{delayed, downgrade}` methods to allow fallibly performing delays/downgrades. Good for debugging code and for applications that can't tolerate panicking
* Worked on the panics used within `.delayed()` & `.downgrade()` with the aim of reducing their code bloat as much as possible, thereby making the compiler eagerly inline the actual function's body without including all the panicking/formatting/generic machinery within it
* Removed `#[inline]` attributes from many frequently-called functions. This aims to counteract code bloat, as the `#[inline]` attribute has large compile time costs ([reference](https://doc.rust-lang.org/reference/attributes/codegen.html#the-inline-attribute), [rust core team member](https://internals.rust-lang.org/t/when-should-i-use-inline/598/3?u=wisagan))

### DIFF
--- a/timely/src/dataflow/operators/capability.rs
+++ b/timely/src/dataflow/operators/capability.rs
@@ -21,7 +21,7 @@
 //! An operator should not hand its capabilities to some other operator. In the future, we should
 //! probably bind capabilities more strongly to a specific operator and output.
 
-use std::ops::Deref;
+use std::{borrow, error::Error, fmt::Display, ops::Deref};
 use std::rc::Rc;
 use std::cell::RefCell;
 use std::fmt::{self, Debug};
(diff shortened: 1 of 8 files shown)


frankmcsherry:
I think this is mostly good, except the removal of the `inline` attribute. Without this attribute, compiled code may not present the Rust code that would allow for inlining, *preventing* the inlining short of something like LTO. In this case, `#[inline(always)]` is probably appropriate, as each call is just a trampoline to another call.

Kixiron:
Generic functions are always valid for inlining regardless of attributes, but the inking attribute is highly detrimental to compile times as it makes a separate copy of each function at *every invocation site*. The inline attribute only affects cross crate inlining in debug mode, which is a mode I'd argue should prioritize compile times above fairly minor performance impacts. Rustc and LLVM are really smart and will inline trivial things like this, the attribute just hurts users. If debug perf is a real issue for people (which it would have been before this), they can use something like ThinLTO to get incremental LTO

frankmcsherry:
Right, but none of the `PartialOrder` implementations are generic. These are things like `u64::less_equal` which we do want to inline to be `<=`.

Kixiron:
Ah, gotcha. I'll add those back to the macro, my bad

frankmcsherry:
Thanks!

PR #388, opened 2021-06-01
add Antichain::into_elements

umanwizard:
This can be useful if one wants to do something with the elements of an antichain without needing to clone them.


frankmcsherry (approved):
Seems legit. tyvm!

PR #389, opened 2021-06-02
Antichain helper implementations

frankmcsherry:
A few convenience implementations for `Antichain<T>` that makes it a bit more like an invariant preserving wrapper for `Vec<T>`, allowing conventional inspection of the vector (e.g. `Deref`, and `Into<Vec<T>>` implementation).

cc: @umanwizard


umanwizard:
Thanks!

PR #391, opened 2021-06-05
Fixes 390

### LINKED ISSUE #390: builder_rc::OperatorBuilder lacks shape methods
`builder_rc::OperatorBuilder` is missing the `.shape()` method which means users can't get the number of inputs or outputs for the operator

frankmcsherry:
Fixes #390 

cc @Kixiron


Kixiron:
Perfect, thanks!

PR #396, opened 2021-06-12
Fix #395: Add next_count returning the notification count

### LINKED ISSUE #395: Notificator notification count is always 1
The notificator's [`for_each`](https://github.com/TimelyDataflow/timely-dataflow/blob/master/timely/src/dataflow/operators/generic/notificator.rs#L82) function claims to invoke the passed function with the parameters `cap, count, notificator`, where count represents how many times `notify_at` has been called for the capability. In reality, it always returns 1.

antiguru:
Fixes #395 


antiguru:
Ok, the fix isn't correct yet, workin on it!

frankmcsherry (approved):
This looks great; thanks for noticing and fixing the issue!

Issue #394, opened 2021-06-11
Lazy memory allocation in channel pushers.


ryzhyk:
We have an application that builds a very large dataflow graph (thousands of operators) and creates hundreds of concurrently running instances of this graph.  Heap profiling showed that the application wastes a lot of memory in channel pushers, which maintain `Message::default_length` entries even when the system is idle with no outstanding messages.

I understand that this is a performance optimization that reduces the number of `malloc` calls, but the memory overhead was not acceptable for our use case, so I created a patch that changes the allocation policy to only allocate channel memory (in chunks of `Message::default_length` entries) when there are messages to send and to deallocate it when there are no messages, reducing the memory footprint to 0 in idle state:
https://github.com/ddlog-dev/timely-dataflow/commit/a94e1763756d01d938d2e0b0ec91c8fd00fc130d

@frankmcsherry , any chance you'd accept a cleaned up version of this patch as a PR?  I realize that it caters for a somewhat idiosyncratic use case, but it doesn't seem to cause a measurable slow down in my experiments and it does save memory.

Kixiron:
If this being the default behavior is a deal breaker it can always be gated either behind a feature flag or behind a timely `Config` option

frankmcsherry:
Sorry for the delay, I was out /  am overwhelmed. :)

The thing you describe sounds like an issue we should fix. If you have a patch, we could totally look at it. It would be a performance regression for some use cases (those that fill these buffers in steady state). Do you have a take on whether your graphs 1. alternate between filling the buffers and then idling on empty buffers, or 2. don't fill the buffers and so a smaller default / adaptive approach would be better, or maybe a bit of both?

frankmcsherry:
Another option, though idk for certain: this could just be a different pact implementation. `Exchange` has an implementation, but it doesn't have to be the pact that you use (we could have both implementations, and users could choose between them as appropriate). That .. only really makes sense if the exchange channel is where the empty buffers languish, rather than at other moments in the dataflow.

frankmcsherry:
There's also, and I'm not sure I want to be that guy but, the possibility that making hundreds of independent copies of a large dataflow graph is just an anti-pattern in TD. I apologize if you have before, but can you explain why this can't be one copy of the large dataflow graph, with keys on the data to indicate which logical instance they belong to?

ryzhyk:
> It would be a performance regression for some use cases (those that fill these buffers in steady state)

[...]

PR #397, opened 2021-06-25
Lazy-allocate pusher memory.

ryzhyk:
The current implementation of channel pushers preallocates
`Message::default_length` entries per pusher.  For very large graphs this
adds up to a lot of memory being allocated even when the system is idle
with no outstanding messages.  This patch changes the allocation policy
to only allocate channel memory when there are messages to send and to
deallocate it when there are no messages, reducing the memory footprint
to 0 in idle state at the cost of some potential slow-down due to a
larger number of allocations.

See #394.


frankmcsherry:
A few quick thoughts:

1. Lazy allocation seems great.
2. The approach to de-allocation will de-allocate on every call to `Message::push_at`, rather than at the end of a stream of messages. The `Push` trait that gets used everywhere communicates "I have no more data" with a `None` message, and de-allocating on that seems better. So, once every burst of data, rather than once for every message.
3. The above approach might mean that this can be implemented by e.g. `Exchange` and then just be another `pact` rather than a system-wide change to resource management. If nothing else, it's probably worth tracking down where these buffers get stashed.

ryzhyk:
Thanks for the feedback. 

> So, once every burst of data, rather than once for every message.

Makes sense!  Instead of always deallocating inside `push_at`, I will modify pusher implementations that maintain internal buffers (`Exchange` and `Tee`) to deallocate the buffer on a `None` message.

> The above approach might mean that this can be implemented by e.g. Exchange and then just be another pact rather than a system-wide change to resource management.

Sorry, I'm not sure how to implement this.  Are you suggesting that `pushers::Exchange` should be configurable to optionally deallocate at the end of a burst, and then we create a pact similar to `ExchangePact` but with this option enabled?  If so, how do I get timely to use this new pact when constructing the dataflow?

frankmcsherry:
Re-reading things, I think the `Exchange` implementation is missing an opportunity to blank out its vector on lines 44-46, where for single-worker instantiations it just passes the message through (including a `None` indicating a flush).

ryzhyk:
I thought in this case the buffer would never be allocated in the first place?

frankmcsherry:
Ah good point. That may be true!

ryzhyk:
I confirmed using out internal benchmarks that the fixed up version that only deallocates on `None` yields memory savings comparable to the version that aggressively deallocated on every push.

@frankmcsherry, do you see this being merged as is or do you feel we should work on making this an optional pact?

frankmcsherry:
I'd need to think a bit before merging it as is, due to the impact on others. If it were an optional pact, it would be easy to accept. On the other hand, I'm not sure that the `Tee` stuff is as easy to make optional, which is too bad. I do like the idea, and we did something like this in Naiad, but I do want to shake out the performance penalty before landing it.

PR #398, opened 2021-07-09
communication: Add zero-copy process configuration

antiguru:
Currently, Timely's communication layer supports a zero-copy process mode,
but it's not exported as part of the `Config` enum. This change adds it
as another alternative to make it simpler for clients to use it.


frankmcsherry (approved):
Seems reasonable to me!

PR #401, opened 2021-07-14
async: relax requirements of supplied streams

petrosagg:
If the user has a stream that is `Unpin` we shouldn't force them to put it in a pinned box


frankmcsherry (approved):

frankmcsherry:
Looks good, and thanks for the explanation!

PR #403, opened 2021-07-15
Base message default length off element size

antiguru:
Instead of defaulting to 1024 elements, use the size in memory of the data
type to determine the default length. With this change, the number of
elements is based off a buffer of 8KiB to match the current behavior for
1024 8-byte elements.

### DIFF
--- a/logging/src/lib.rs
+++ b/logging/src/lib.rs
@@ -123,7 +123,7 @@ impl<T, E: Clone> Logger<T, E> {
             time,
             offset,
             action: action,
-            buffer: Vec::with_capacity(1024),
+            buffer: Vec::new(),
         };
         let inner = Rc::new(RefCell::new(inner));
         Logger { inner }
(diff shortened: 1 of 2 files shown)


timely/src/dataflow/channels/mod.rs
```diff
@@ -28,7 +28,12 @@ pub struct Message<T, D> {
 impl<T, D> Message<T, D> {
     /// Default buffer size.
     pub fn default_length() -> usize {
-        1024
+        const MESSAGE_BUFFER_SIZE: usize = 1 << 13;
+        match std::mem::size_of::<D>() {
+            0 => MESSAGE_BUFFER_SIZE,
```
Kixiron, line 33: If it's a zst you may as well just use `usize::MAX`, vectors of zsts don't allocate anyways

antiguru: You're right! Changed the PR!

antiguru: As it turns out, `usize::MAX` leads to overflows in Timely's `to_stream` operator.

antiguru:
The PR now also modifies the allocation behavior for the logging crate, which suffers from a similar issue.

The `TimelyEvent` enum currently has a size of ~72 bytes, which turns into `(Duration, usize, TimelyEvent)` into ~96 bytes, resulting in allocations of ~90KiB per log buffer. (I did the measurements yesterday and might be off by 8 bytes or so.)

logging/src/lib.rs
```diff
@@ -168,6 +168,25 @@ impl<T, E: Clone> Logger<T, E> {
...
+    // TODO: This fn is not const because it cannot depend on non-Sized generic parameters
+    fn buffer_capacity() -> usize {
+        let size =  ::std::mem::size_of::<(Duration, E, T)>();
+        if size == 0 {
+            Self::BUFFER_SIZE_BYTES
+        } else if size <= Self::BUFFER_SIZE_BYTES {
+            Self::BUFFER_SIZE_BYTES / size
+        } else {
+            1
+        }
```
frankmcsherry, line 187: very minor nit, but is there a reason to do this as if statements, and in the timely crate as a match? This version is clearer to me (unfamiliar with Rust's match guarding logic), but if there is a reason let me know! Can fix or not fix; very minor.

frankmcsherry (approved):
LGTM, thanks!

PR #404, opened 2021-07-15
replay: work with rust streams instead of event iterators

petrosagg:
This patch makes the Replay operator accept a list of streams rather than a list of event iterators. The streams are consumed with a sync activator wrapper in a Waker in order to re-schedule the operator automatically when new data becomes available.

### DIFF
--- a/kafkaesque/Cargo.toml
+++ b/kafkaesque/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2018"
 clap="*"
 abomonation="0.7"
 timely = { path = "../timely" }
+futures = "0.3"
 
 [dependencies.rdkafka]
 version = "0.23.0"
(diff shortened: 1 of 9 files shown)


frankmcsherry:
Thanks! Can you say more about why this should land? It seems to be strictly worse for `EventIterator` types that can produced references and do not need to clone their contents. Am I missing something there?

petrosagg:
The reason for the async stream part is based on the assumption that most of the time when you replay a stream you're getting it from the network and presumably you're using something like tokio (may or may not be true).

On the cloning, I noticed that the current implementation clones both of its arguments unconditionally https://github.com/TimelyDataflow/timely-dataflow/blob/master/timely/src/dataflow/operators/capture/replay.rs#L96-L99 which is why I thought it's ok to make it receive owned items. It is not 100% the same, as the previous version cloned all the `D`s and this one clones the vector itself, but if you think the previous behaviour should be kept I can make a ReplayAsync version of the trait that works over async streams

PR #407, opened 2021-07-22
logging: Store logger in Rc

antiguru:
The idea of this PR is to reduce the size of various structs that contain `Logger`s. For example, the size of `LogPusher` drops from 120 bytes to 64 bytes, which can lead to a noticeable reduction in memory with large dataflows and many workers.

On the other hand, this change introduces API-incompatibilities due to the change of the logging type from `Logger` to `Rc<RefCell<Logger>>`.


antiguru:
As it turns out, this is not a good change. Using the logging API would be significantly harder than it is right now.

PR #408, opened 2021-07-22
Encapsulate logger state in shared structure

antiguru:
Encapsulate the state of a logger in a `Rc<RefCell<...>>` to reduce the amount of state in all places where loggers are retained. This reduces the memory requirements for some structures considerable. `LogPusher` changes from 120 to 72 bytes, and similar reductions can be expected in other places.

This change does not alter the public API. It should not have significant performance implications as logging would already dereference a pointer to a shared buffer and a shared action function, which are now behind a single reference cell.

### DIFF
--- a/logging/src/lib.rs
+++ b/logging/src/lib.rs
@@ -87,24 +87,29 @@ impl<Id> Flush for Registry<Id> {
 }
 
 /// A buffering logger.
+#[derive(Debug)]
 pub struct Logger<T, E> {
-    id:     E,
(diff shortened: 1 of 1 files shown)


frankmcsherry:
Thanks! This seems good, but we may revisit in the future. It seems to rule out "one buffer, multiple actions", which may be handy if we wanted e.g. to change the logging to allow compaction or filtering in-place. But, I think for the moment it's all good!

PR #409, opened 2021-07-30
Disable LTO

antiguru:
Disable LTO for timely-dataflow because it causes erratic runtime performance, at least in Rust 1.53.


frankmcsherry (approved):
Seems good.

PR #410, opened 2021-07-31
Update link to GitHub repo

tisonkun:
I'm unsure whether it is intend to be so, but it looks like a typo to me.


frankmcsherry:
The change is good! It used to be my repo, but it was shifted over when I went corporate.

PR #414, opened 2021-08-04
Implement broadcast with custom Pact

antiguru:
Currently, the broadcast operator duplicates the data for each peer and
uses the regular exchange infrastructure to send the duplicates to each
peer, such that each peer will receive a complete copy of the stream.

This is problematic as on the sending worker as many copies of the data
as there are peers need to exist. Also, the operator is implemented as
three operators, which needs additional copies of the data.

With this change, a specialized parallelization contract handles
distributing the data. Copies are avoided by passing reference-counted
batches of data to the peer, which means that all peers get to see the
same data, at least for process-local communication.

### DIFF
--- a/timely/src/dataflow/operators/broadcast.rs
+++ b/timely/src/dataflow/operators/broadcast.rs
@@ -1,8 +1,16 @@
 //! Broadcast records to all workers.
 
+use std::marker::PhantomData;
+use std::sync::Arc;
+
+use crate::communication::{Pull, Push};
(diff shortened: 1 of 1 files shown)


antiguru:
PS: This needs a test!

antiguru:
As it turns out, this messes with Timely's progress tracking. Progress tracking assumes that there is a 1:1 relationship of data going into a PACT and coming out again. This implementation breaks the contract by internally duplicating the messages. Might take a look if this is something that can be fixed with reasonable effort!

PR #416, opened 2021-08-18
Exchange: bench, do not preallocate buffers

antiguru:
This PR:
* Changes the exchange behavior to only allocate push buffers when encountering data.
* Adds a benchmark to compare exchange changes.

I compared the current implementation versus the changes in this PR, and the benchmark doesn't report significant changes in performance. Obviously, this PR encodes a policy change from eagerly allocated buffers to allocating them on-demand, which could have an impact on memory utilization and processing latency.

As explained on the commit message, the cost of allocating buffers exists today, only that it is now during initial construction and with the default process allocator after push operations. We shift those to allocations before the fist buffer push, which means that the absolute number of allocations stays the same. Regarding latency, we need to check before pushing to a buffer whether it has the right capacity, which might add to latency. However, I'd hope that the compiler can fold this into the subsequent test whether the buffer is full, so I wouldn't expect much of a latency impact either.


antiguru:
Benchmark results temporarily at https://dev.antiguru.de/criterion/report/

antiguru:
@Kixiron, @ryzhyk, this PR provides an `EagerExchange`, which should correspond roughly to what you proposed in #397, but not the `Tee` changes. It'd be great if you could test this change with your benchmark.

ryzhyk:
> @Kixiron, @ryzhyk, this PR provides an `EagerExchange`, which should correspond roughly to what you proposed in #397, but not the `Tee` changes. It'd be great if you could test this change with your benchmark.

@antiguru, very sorry, I missed your message, will give it a try and let you know.

ryzhyk:
@antiguru , I took a close look at the PR.  Unfortunately, trying it out is not as easy as just switching to a different timely branch; I'd have to modify our code that constructs the dataflow graph to apply the new pacts to every operator.  I apologize, but unfortunately I won't have the cycles to work on this now :(

frankmcsherry (approved):
Looks good to me! If you don't mind, adding comments around the two new allocation blocks would make sure we don't forget why we are doing this, and prevents future me from just copying the `push()` fragment without the allocation test.

PR #417, opened 2021-08-24
Fix error message on capability downgrade

aljoscha:


timely/src/dataflow/operators/capability.rs
```diff
@@ -478,7 +478,7 @@ impl<T: Timestamp> CapabilitySet<T> {
             // we outline it
             panic!(
                 "Attempted to downgrade a CapabilitySet with a frontier containing elements \
-                which where were not not `less_equal` than all elements within the set"
+                which where were not beyond all elements within the set"
```
frankmcsherry, line 481: Nit, but this might also be a good time to change this text. It should "... containing an element that was not beyond an element within the set".

frankmcsherry (approved):
Looks good to me. If you have a moment to tweak the `CapabilitySet` error text it was not right and remains not right, but that could also be fixed later on.

aljoscha:
> Looks good to me. If you have a moment to tweak the CapabilitySet error text it was not right and remains not right, but that could also be fixed later on.

Sure! I pushed the change.

frankmcsherry:
Thanks!

PR #419, opened 2021-09-10
Explicitly implement Clone for Antichain

frankmcsherry:
Ignore the incorrect branch name. Explicitly implements `Clone` so that we have access to a `clone_from` that doesn't just call `clone()`. Rust does not derive `clone_from` implementations: https://github.com/rust-lang/rust/pull/27939.


antiguru (approved):
LGTM, thanks a bunch!

PR #420, opened 2021-09-16
Logging: Do not flush on drop, only on last copy

antiguru:
The current implementation of the logger flushes when dropping the outer
logger, which contains an Rc to the inner logger. This causes excessive
flushes for clients obtaining loggers frequently. With this change, the
flush is only triggered when dropping the inner logger, i.e. when the last
copy of the Rc goes out of scope.


frankmcsherry (approved):
I think this looks good, but would you mind updating the comment (doccomment? weird.) on the implementation. I believe this PR intends to correct its concern.

antiguru:
Comment fixed! Thanks for the review!

PR #421, opened 2021-09-16
Avoid always flushing logs in worker

antiguru:
Currently, we flush on Park and after the `step_or_park` finishes. With
this change, we only explicitly flush on Park.

Note that this might delay some log messages for long park intervals.


frankmcsherry (commented):
I'm not sure about this change. You can run timely in a non-parking configuration, and this change would (I think) mean that we would sit on log data indefinitely in that config. I get the intent that we should allow logging to flush less frequently, but I think it is more subtle than this edit.

PR #429, opened 2021-11-02
Safe(r) minting of `CapabilityRef`s

petrosagg:
### Background

As part of progress tracking timely keeps track of three kinds of timestamp frequencies for each operator. Timestamps at the inputs (the `consumeds` part of `SharedProgress`), internal timestamps (the `internals` part of `SharedProgress`) and timestamps at the output (the `produceds` part of `SharedProgress`). A timely operator must carefully manipulate those frequencies in order to let timely know at what times data might be produced from the operator. An operator can misuse these frequencies if for example it clears a `consumed` timestamp `t1` without adding it to its `internal` timestamps and later decides that it wants to produce data at `t1`.

With the exception of the raw operator builder timely provides a safe API that ensures timestamp frequency counts are correct via the `CapabilityRef` and `Capability` types. These act as a witness of the existence of a `consumed` or `internal` timestamp respectively.

### CapabilityRef misuse

The `Capability` type which represents an `internal` timestamp contains within it a shared reference to the internal timestamp frequency counts and it automatically subtracts its timestamp from the counts on `Drop`. This allows users to keep a `Capability` around in whatever way they wish and have the internal operator timestamp counts follow accordingly.

However, the same is not true for the `CapabilityRef` type. When a `CapabilityRef` is minted its timestamp is *immediately* subtracted from `consumed` and provided to the user. This is a problem because if the operator logic held onto the `CapabilityRef` then it could later produce data that is behind the frontier.

In practice, holding onto a `CapabilityRef` across operator invocations is quite difficult as the user will be tasked with storing an input handle and its produced `CapabilityRef` (which is lifetimed) at the same time. Self-referencial structs are notoriously difficult (but not impossible nor unsound) to do in Rust and so synchronous operators don't usually face this problem.

This situation becomes trivially possible with an async operator. In async rust, where the compiler is tasked with generating a struct that captures the stack of a future at any given yield point, making this self-referencial is easy. A user simply needs to obtain a `CapabilityRef` from an input handle and then `await` on something. When this yield point is reached the current stack will be preserved, keeping the `CapabilityRef` alive across timely invocations.

### Solution in this PR

The solution this PR implements is to make `CapabilityRef` behave in the same way as a `Capability`. Instead of relying on the difficulty of self-referencial structs in Rust this PR adds an extra guard field in `CapabilityRef` that when dropped it will update the `consumed` timestamps of the operator accordingly. This allows users to keep `CapabilityRef`s around for as long as they wish.

### DIFF
--- a/timely/src/dataflow/channels/pullers/counter.rs
+++ b/timely/src/dataflow/channels/pullers/counter.rs
@@ -15,14 +15,36 @@ pub struct Counter<T: Ord+Clone+'static, D, P: Pull<BundleCore<T, D>>> {
     phantom: ::std::marker::PhantomData<D>,
 }
 
+/// A guard type that updates the change batch counts on drop
+pub struct ConsumedGuard<'a, T: Ord + Clone + 'static> {
+    consumed: &'a Rc<RefCell<ChangeBatch<T>>>,
+    time: Option<T>,
+    len: usize,
+}
+
+impl<'a, T:Ord+Clone+'static> Drop for ConsumedGuard<'a, T> {
+    fn drop(&mut self) {
+        self.consumed.borrow_mut().update(self.time.take().unwrap(), self.len as i64);
+    }
+}
+
 impl<T:Ord+Clone+'static, D: Container, P: Pull<BundleCore<T, D>>> Counter<T, D, P> {
     /// Retrieves the next timestamp and batch of data.
     #[inline]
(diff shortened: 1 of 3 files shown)


petrosagg:
I just built materialize with this change to get further confidence that it is correct. Tests passed :)  https://github.com/MaterializeInc/materialize/pull/9023

frankmcsherry:
Looks good; thanks!

PR #431, opened 2021-11-03
operators: expose the scope of generic operators

petrosagg:


antiguru:
I can't start a review, so my comments here: The only change I'd do is to return `&G` instead of eagerly cloning, but otherwise it seems fine. I wonder, though, don't all the places that construct a builder already have access to the scope? They'd at least need it to create the builder I think.

petrosagg:
Good point on returning `&G`, fixed. I copied the method from Stream without too much thought.

Regarding what's the use for this kind of function, it's for when you want to write an extension trait on the builder where you don't have access to the arguments that it got constructed with. This is where I needed this concretely and without this I was forced to add a scope parameter: https://github.com/MaterializeInc/materialize/blob/main/src/persist/src/operators/async_ext.rs#L115

petrosagg:
Closing this as it's no longer needed

PR #434, opened 2021-11-20
exchange: De-duplicate partitioning logic

antiguru:
Currently, the partitioning logic has two separate branches for
power-of-two peers using bit masking and mod for all other configurations.
The branches duplicate the logic to distribute the data to the pushers.
This change moves the common functionality into a separate function.

We force inlining of the partition function to avoid a performance penalty.

### DIFF
--- a/timely/src/dataflow/channels/pushers/exchange.rs
+++ b/timely/src/dataflow/channels/pushers/exchange.rs
@@ -35,6 +35,25 @@ impl<T: Clone, D, P: Push<Bundle<T, D>>, H: FnMut(&T, &D)->u64>  Exchange<T, D, 
             }
         }
     }
+
+    /// Partition data according to an index function
+    #[inline(always)]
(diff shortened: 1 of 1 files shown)


timely/src/dataflow/channels/pushers/exchange.rs
```diff
@@ -35,6 +35,25 @@ impl<T: Clone, D, P: Push<Bundle<T, D>>, H: FnMut(&T, &D)->u64>  Exchange<T, D,
             }
         }
     }
+
+    /// Partition data according to an index function
+    #[inline(always)]
+    fn partition<F: Fn(u64) -> usize>(&mut self, time: &T, data: &mut Vec<D>, func: F) {
```
antiguru, line 41: Rename to `push_partitioned`

frankmcsherry:
Looks great with naming nit fixed; thanks very much!

PR #435, opened 2021-11-20
Exchange sort

antiguru:
This is a draft to show how Timely could capture the exchange pattern, sort it, and reduce the cache misses for pushers.

Currently, this results in a significant performance regression for a low number of peers, but I'll try to test it on a machine with more workers.


antiguru:
Not pursuing this at the moment.

PR #436, opened 2021-11-30
Update rand to 0.8

antiguru:
Update the dependency on `rand` to version 0.8.


frankmcsherry (approved):
These changes all look good, thank you!

PR #438, opened 2021-12-02
Avoid preallocating buffers in buffer, tee

antiguru:
Instead of allocating buffers at construction time, allocate buffers
before pushing or cloning. This adds additional instructions on the fast
path for input/buffer/exchange/tee, but seems to be justified by the
memory savings.

We expect a linear amount of memory savings per operator in the number of workers. For non-serializing allocators, this memory saving will persist during the lifetime of the instance, for returning (serializing) allocators, the memory savings will remain until the first data is sent.

The exchange benchmark does not report a change in performance.

The main question that we'd need to answer is whether we're comfortable with `Message::push_at` not ensuring that the returned object is allocated. With current Timely, this moves the burden of ensuring allocations to clients of `push_at`, which is not great. Considering #426, this problem would disappear (somewhat) because containers would be responsible for their own buffer management. We could delay this PR until after #426 lands

### DIFF
--- a/timely/src/dataflow/channels/mod.rs
+++ b/timely/src/dataflow/channels/mod.rs
@@ -47,25 +47,12 @@ impl<T, D> Message<T, D> {
     }
 
     /// Forms a message, and pushes contents at `pusher`. Replaces `buffer` with what the pusher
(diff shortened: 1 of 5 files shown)


antiguru:
Would need to be rewritten.

PR #439, opened 2021-12-02
Push filter to upstream operator

antiguru:
Allow operators to push filter expressions to the upstream `Tee`.

This can reduce the amount of data copied within tees, but care needs to be taken to ensure the progress tracking invariants.

The intention is not to land this change but rather have a discussion if this is the right direction for such a change. I don't think the progress tracking wiring is particularly elegant, but seems to do the job. Maybe we could find a better abstraction here. This PR also breaks a public interface, the `Stream::connect_to` function. I'm not sure if there are any users outside of Timely that call this function.

### DIFF
--- a/timely/src/dataflow/operators/filter.rs
+++ b/timely/src/dataflow/operators/filter.rs
@@ -1,9 +1,10 @@
 //! Filters a stream by a predicate.
 
 use crate::Data;
+use crate::communication::message::RefOrMut;
 use crate::dataflow::channels::pact::Pipeline;
 use crate::dataflow::{Stream, Scope};
-use crate::dataflow::operators::generic::operator::Operator;
+use crate::dataflow::operators::generic::builder_rc::OperatorBuilder;
 
 /// Extension trait for filtering.
 pub trait Filter<D: Data> {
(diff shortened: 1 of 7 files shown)


timely/src/dataflow/operators/filter.rs
```diff
@@ -24,15 +25,47 @@ pub trait Filter<D: Data> {
...
-                data.swap(&mut vector);
-                vector.retain(|x| predicate(x));
-                if !vector.is_empty() {
-                    output.session(&time).give_vec(&mut vector);
+        let mut builder = OperatorBuilder::new("Filter".to_owned(), self.scope());
+
+        let mut input = builder.new_input_filter(self, Pipeline, vec![], move |data, buffer| {
+            match data {
+                RefOrMut::Ref(data) => {
+                    buffer.extend(data.into_iter().filter(|x| predicate(*x)).cloned())
```
Kixiron, line 33: Fairly unrelated, but it'd be nice if operators could pass on references instead of requiring cloning, that could remove a lot of fairly redundant clones

antiguru: This might do what you suggested: https://github.com/TimelyDataflow/timely-dataflow/pull/471

PR #440, opened 2021-12-03
builder_ref: first step towards ref based operator builder

petrosagg:
This commit is the first step towards building an operator builder that
doesn't rely on holding on to `Rc<RefCell<ChangeBatch<T>>`.

In order to do so `builder_ref::OperatorBuilder` encodes the current
shape of the operator in the system by consuming self and producing a
new builder type every time an input or output is added.

When the time comes to build the operator, the system requires the user
to supply a closure whose type signature matches exactly the inputs and
outputs have have been previously declared.

This PR only adds the typestate machinery to
`builder_ref::OperatorBuilder`. A subsequent PR will work on removing
all the shared refcounted change batches since they now have a single
owner. The reason this isn't included in this PR is to keep it at a
reasonable size since many auxiliary types like the handles will need to
be adjusted to work with references.

### DIFF
--- a/timely/src/dataflow/operators/generic/builder_ref.rs
+++ b/timely/src/dataflow/operators/generic/builder_ref.rs
@@ -1,5 +1,229 @@
 //! Types to build operators with general shapes.
 
-// TODO: Analogue to builder_rc.rs, which instead provides wrappers that only borrow ChangeBatch<T>
-//       buffers, rather than wrapping them with Rc<RefCell<_>>. This removes dereferences from the
-//       common path of having nothing to do, and could plausibly make it a bit faster to do nothing.
+use std::rc::Rc;
+use std::cell::RefCell;
+
+use crate::Data;
+use crate::dataflow::channels::pact::ParallelizationContract;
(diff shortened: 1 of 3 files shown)


timely/src/dataflow/operators/generic/builder_ref.rs
```diff
@@ -1,5 +1,229 @@
...
+use crate::logging::TimelyLogger as Logger;
+use crate::progress::{ChangeBatch, Timestamp};
+use crate::progress::operate::SharedProgress;
+use crate::progress::frontier::{Antichain, MutableAntichain};
+
+use super::builder_raw::OperatorBuilder as OperatorBuilderRaw;
+
+/// Builds operators with generic shape.
+#[derive(Debug)]
+pub struct OperatorBuilder<G: Scope, Inputs, Outputs> {
```
petrosagg, line 26: Everything in the implementation is currently identical to `builder_rc::OperatorBuilder`. The difference is in the correct tracking and accumulation of the `Inputs` and `Outputs` types.

PR #441, opened 2021-12-09
store bare `SharedProgress` in operators

petrosagg:
Timely only needs temporary access to an operator's SharedProgress data structure so this patch makes it so it is fully owned by operators and is exposed to the outside world through a temporary mutable reference.

Using the `examples/barrier` microbenchmark this leads to a 5% performance increase on my machine:

```bash
[petrosagg@doggo:~/projects/timely-dataflow/timely]$ time ./barrier-master 10000000

real	0m14.548s
user	0m14.545s
sys	0m0.003s

[petrosagg@doggo:~/projects/timely-dataflow/timely]$ time ./barrier-bareprogress 10000000

real	0m13.755s
user	0m13.755s
sys	0m0.001s
```


timely/src/progress/subgraph.rs
```diff
@@ -463,22 +461,26 @@ where
...
-                    .borrow_mut()
-                    .frontiers[port]
-                    .update(time, diff);
+                if let Some(operator) = child.operator.as_mut() {
+                    operator.shared_progress_mut().frontiers[port].update(time, diff);
+                } else {
+                    // If the operator is closed and we are reporting progress at it, something has surely gone wrong.
+                    println!("Operator prematurely shut down: {}", child.name);
+                    println!("  {:?}", child.notify);
+                    // panic!();
```
petrosagg, line 477: I was able to figure out what is going on here. This code patch definitely triggers even though an operator might be shut down. Ignoring it makes the tests pass but from the surrounding code and comments it sounds like that when an operator is shut down we should not push more progress info to it.

The test that makes this code path trigger is `dataflow::operators::enterleave::test::test_nested`

frankmcsherry: I'll take a peek at this. The PR looks good other than commenting out the asserts, but it's great that you are observing them firing because .. should be immediately actionable!

petrosagg: Oh great! Also, above I meant to write "I wasn't able to figure out". I've no idea why this change triggers it :)

petrosagg: @frankmcsherry I debugged this today. In the subgraph handling we are creating a dummy operator as the 0th child that represent the external scope and we create a `PerOperatorState::empty()` to track it.

Previously that struct held onto a synthesized `Rc<RefCell<SharedProgress<T>>>` that didn't come from any operator and was fabricated on the spot. Since this PR moves the `SharedProgress` into the operators themelves this left these dummy operator states without a `SharedProgress` struct.

I fixed it by creating an `EmptyOperator` struct that implements `Operate` and `Schedule` that matches the properties that `PerOperatorState::empty()` already had, which now serves as the holder of `SharedProgress` for the representative.

I isolated the fix to the latest commit.

timely/src/progress/subgraph.rs
```diff
@@ -689,17 +690,7 @@ impl<T: Timestamp> PerOperatorState<T> {
...
-        else {
-
-            // If the operator is closed and we are reporting progress at it, something has surely gone wrong.
-            if self.shared_progress.borrow_mut().frontiers.iter_mut().any(|x| !x.is_empty()) {
-                println!("Operator prematurely shut down: {}", self.name);
-                println!("  {:?}", self.notify);
-                println!("  {:?}", self.shared_progress.borrow_mut().frontiers);
-                panic!();
-            }
-
```
frankmcsherry, line 702: Does this still need to be deleted with the proxy `EmptyOperator`?

timely/src/progress/subgraph.rs
```diff
@@ -785,3 +780,52 @@ impl<T: Timestamp> Drop for PerOperatorState<T> {
...
+    }
+    fn inputs(&self)  -> usize {
+        self.inputs
+    }
+    fn outputs(&self) -> usize {
+        self.outputs
+    }
+    fn get_internal_summary(&mut self) -> Vec<Vec<Antichain<T::Summary>>> {
+        Vec::new()
+    }
```
frankmcsherry, line 813: I wonder (I can try it) whether this should perhaps be a `panic!` in order to avoid miscommunicating. Afaiu, this struct exists only to house the `SharedProgress<T>` struct, and doing anything else with it seems scary.

frankmcsherry (commented):
This looks good. One thing I'd like to understand is whether this is the most efficient approach, as opposed to the opposite where the parent holds the `SharedProgress<T>` and passes it as an argument to `schedule`. In that world, there is no `Box<Operate>::shared_progress_mut()` being called, always the right type is at hand, etc.

If you happen to know off the top of your head if/why that is bad I'd be interested, and if not I'm tempted to try it out myself and report back!

frankmcsherry:
Maybe a riff on the above, unrelated to this PR: `Operate` was whipped together a long time ago, and it seems now that perhaps not all of its methods should be "virtual". We have the `PerOperatorState` in part to lock down bits of the `Box<Operate>` that we don't want to have to revisit. Potentially the only thing that should be boxed at all is `schedule`, though .. that's probably wrong.

timely/src/progress/subgraph.rs
petrosagg: We still check it, it is just moved further up https://github.com/TimelyDataflow/timely-dataflow/pull/441/files#diff-ac6e819a79d1d2170b82aee757936570e29d586f825d119f8b1092ee8df187f4R474

frankmcsherry:
Well, a brief inspection suggests that this is all complicated by `trait Schedule` having the `schedule` method, but no `T` type parameter. I think it should ultimately be better to call in to operators with arguments once rather than ask them to get handles to their results, but ... we'll have to see what the path to that looks like.

timely/src/progress/subgraph.rs
petrosagg: Good point, I added an `unreachable!()` call with an appropriate message

frankmcsherry:
I whipped up a version that does the "parent owned" `SharedProgress`, and .. I can't see all that much difference between the three branches. Would you say your runtimes were representative, @petrosagg, or might they have been kinda random too?
```
MASTER: 
cargo run --release --example barrier -- 10000000  13.20s user 0.07s system 98% cpu 13.405 total
cargo run --release --example barrier -- 10000000  12.85s user 0.06s system 98% cpu 13.057 total
cargo run --release --example barrier -- 10000000  12.91s user 0.06s system 98% cpu 13.116 total

PETROS: 
cargo run --release --example barrier -- 10000000  13.09s user 0.06s system 98% cpu 13.296 total
cargo run --release --example barrier -- 10000000  13.09s user 0.06s system 98% cpu 13.291 total
cargo run --release --example barrier -- 10000000  13.05s user 0.06s system 98% cpu 13.249 total

FRANK:
cargo run --release --example barrier -- 10000000  13.05s user 0.08s system 98% cpu 13.322 total
cargo run --release --example barrier -- 10000000  12.70s user 0.06s system 98% cpu 12.892 total
cargo run --release --example barrier -- 10000000  12.95s user 0.07s system 98% cpu 13.183 total
```

frankmcsherry:
I just tried it again bypassing `cargo` just to see .. and `master` seems to be the best at the moment. Feels a bit like there is some other randomness in there (optimization success, idk?). 
```

MASTER:
mcsherry@gallustrate timely-dataflow % time ./target/release/examples/barrier 10000000
./target/release/examples/barrier 10000000  12.38s user 0.01s system 99% cpu 12.403 total
mcsherry@gallustrate timely-dataflow % time ./target/release/examples/barrier 10000000
./target/release/examples/barrier 10000000  12.41s user 0.01s system 99% cpu 12.434 total
mcsherry@gallustrate timely-dataflow % time ./target/release/examples/barrier 10000000
./target/release/examples/barrier 10000000  12.64s user 0.02s system 99% cpu 12.686 total
mcsherry@gallustrate timely-dataflow % 

PETROS:
mcsherry@gallustrate timely-dataflow % time ./target/release/examples/barrier 10000000
./target/release/examples/barrier 10000000  13.07s user 0.01s system 99% cpu 13.098 total
mcsherry@gallustrate timely-dataflow % time ./target/release/examples/barrier 10000000
./target/release/examples/barrier 10000000  13.07s user 0.01s system 99% cpu 13.098 total
mcsherry@gallustrate timely-dataflow % time ./target/release/examples/barrier 10000000
./target/release/examples/barrier 10000000  12.91s user 0.01s system 99% cpu 12.929 total
mcsherry@gallustrate timely-dataflow % 

FRANK:
mcsherry@gallustrate timely-dataflow % time ./target/release/examples/barrier 10000000
./target/release/examples/barrier 10000000  12.49s user 0.01s system 99% cpu 12.511 total
mcsherry@gallustrate timely-dataflow % time ./target/release/examples/barrier 10000000
./target/release/examples/barrier 10000000  12.49s user 0.01s system 99% cpu 12.513 total
mcsherry@gallustrate timely-dataflow % time ./target/release/examples/barrier 10000000
./target/release/examples/barrier 10000000  12.73s user 0.01s system 99% cpu 12.755 total
mcsherry@gallustrate timely-dataflow % 
```

PR #444, opened 2022-01-15
MapCore for containers

antiguru:
An idea on how we could provide a generic map function for containers, where the client's interface does not need to know the specifics of the container type.

This includes a draft on how we could provide a `flat_map` operator, but it is unclear to me whether the recycling of allocations is desirable here.


antiguru:
Not needed anymore.

PR #445, opened 2022-01-21
Allow exchange operator to take FnMut

davidselassie:
Since the underlying `channels::pact::Exchange` already can store an `FnMut` as the hash function, it seems like the exchange operator should be able to support taking one as well. I'm still getting my bearings with Timely and Rust, so if there's a reason for the mismatch, I'd love to learn about why.

FWIW, my use case: for reasons unrelated to Timely, I have a limit on how many workers can gather input from the previous system, so I'd like to round-robin exchange right away with a counter, so additional workers have things to chew on in the early part of the dataflow. I know I could hash the input or generate random numbers currently, but this seemed like it could be supported as well.

Thanks!


frankmcsherry:
Seems good; thanks!

PR #446, opened 2022-01-23
Container contents

antiguru:
Implement abstractions for containers that allow clients to append to a container and drain all its elements. As containers can either reveal owned items (e.g., `Vec`), or references (e.g., `TimelyStack`), the API is generic over a `RefOrOwned` type that can both represent references and owned data.

Why not use the `RefOrMut` type form Timely's communication crate? We could provide access to the elements drained from a `Vec` as a `&mut`, which would allow clients to move out of the reference. However, this requires that the type is `Default`, because we need to leave something behind. Imposing `Default` on `Data` seems a significant change. In addition to this, there is no mechanism to recycle allocations through elements of containers (rather than containers itself), so the reason why `RefOrMut` encapsulates a `&mut` doesn't apply here.

With this abstraction in place, we convert several operators that require access to the contents of a container to generic variants.


container/src/lib.rs
```diff
@@ -40,6 +42,57 @@ pub trait Container: Default + Clone + 'static {
...
+    fn reserve(&mut self, additional: usize) {
+        Vec::reserve(self, additional)
+    }
+
+    fn show(&mut self, element: RefOrMut<Self::Item>) {
+        match element {
+            RefOrMut::Ref(reference) => self.push(reference.clone()),
+            // The following `clone` is annoying. `T` is not `Default`, so we cannot
+            // move out of the mutable reference.
+            RefOrMut::Mut(reference) => self.push(reference.clone()),
```
antiguru, line 75: Ideally we'd move out of the mutable reference, but since `T` is not `Default`, we can't. Adding the `Default` constraint requires to change `communication::Data`, which has undesirable side-effects.

antiguru:
Superseded by current container abstractions.

PR #454, opened 2022-03-07
implement Hash for Antichain<T> when T is Ord+Hash

danhhz:
This allows for e.g. indexing differential Descriptions by their lower
to stitch them together.

### DIFF
--- a/timely/src/progress/frontier.rs
+++ b/timely/src/progress/frontier.rs
@@ -229,6 +229,16 @@ impl<T: Clone> Clone for Antichain<T> {
 
 impl<T: TotalOrder> TotalOrder for Antichain<T> { }
 
+impl<T: Ord+std::hash::Hash> std::hash::Hash for Antichain<T> {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let mut temp = self.elements.iter().collect::<Vec<_>>();
+        temp.sort();
+        for element in temp {
+            element.hash(state);
+        }
+    }
+}
+
 impl<T: PartialOrder> From<Vec<T>> for Antichain<T> {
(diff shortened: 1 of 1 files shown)


timely/src/progress/frontier.rs
```diff
@@ -229,6 +229,16 @@ impl<T: Clone> Clone for Antichain<T> {
 
 impl<T: TotalOrder> TotalOrder for Antichain<T> { }
 
+impl<T: Ord+std::hash::Hash> std::hash::Hash for Antichain<T> {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        let mut temp = self.elements.iter().collect::<Vec<_>>();
```
danhhz, line 234: this temp Vec is unfortunate, should I be doing something different? maybe impl Hash for MutableAntichain (which requires that T be Ord) instead?

danhhz (commented):
@frankmcsherry I double-checked after office hours and it looked like we indeed didn't have this. no hurry at all! we might want something like this eventually, but not soon

timely/src/progress/frontier.rs
frankmcsherry: This seems fine to me, and is in-line with how we do equality. It just means that `Hash` is not super cheap, but that isn't the end of the world imo. A different take is that we should maintain antichains in normal form (sorted, say), which is a different amount of disruptive.

danhhz: Okay then it works for me, too! It's still much faster than the alternative (repeatedly iterating through a list to call Eq). Just wanted to make sure I wasn't missing something.

frankmcsherry:
Looks good, and thanks!

PR #457, opened 2022-03-18
Remove Abomonation for encoding MessageHeader

antiguru:
While this does not remove the dependency on Abomonation, it ensures that
reading and writing MessageHeader structs from the network does not use
Abomonation.


frankmcsherry:
This seems like a good idea.

frankmcsherry (approved):
Seems good to me; shall we merge?

antiguru:
Yup, easy to back out if it's not what we like it to be.

PR #458, opened 2022-03-18
Add option conversions for totally ordered antichains

frankmcsherry:
Antichains over totally ordered elements always contain at most one element, and so can be converted into `Option` types. These implementations make that the responsibility of the implementation here.


benesch (approved):
Exciting stuff!

PR #459, opened 2022-03-18
Antichain `FromIterator` implementations

frankmcsherry:
For both `Antichain` and `MutableAntichain`.

cc: @benesch


benesch (approved):
Seems good!

PR #460, opened 2022-03-18
Remove all Pusher/Puller boxing using GATs

petrosagg:
Currently the `Allocate::allocate` method is hardcoded to return a `(Vec<Box<dyn Push<Message<T>>>>, Box<dyn Pull<Message<T>>>)` pair. Boxing was necessary because the generic `T` parameter was on the method.

Generic associated types is the missing piece to allow the concrete type to be spelled out and since they are right around the corner for stable Rust I went ahead and implemented it. 

This PR removes every single use of a `Box<dyn Puller/Pusher>` instance in the codebase. It is not mergeable yet because it requires the nightly feature flag but we could discuss it and merge it once GATs are released.


timely/src/dataflow/channels/pact.rs
```diff
@@ -14,37 +14,36 @@ use crate::communication::{Push, Pull, Data};
...
 use crate::dataflow::channels::pushers::Exchange as ExchangePusher;
+use crate::dataflow::ScopeParent;
 use super::{BundleCore, Message};
 
 use crate::logging::{TimelyLogger as Logger, MessagesEvent};
-use crate::progress::Timestamp;
 
 /// A `ParallelizationContractCore` allocates paired `Push` and `Pull` implementors.
-pub trait ParallelizationContractCore<T, D> {
+pub trait ParallelizationContractCore<G: ScopeParent, D> {
```
frankmcsherry, line 24: This looks like a regression. Can we leave it as `T`?

petrosagg: Yes, just fixed that

timely/src/progress/broadcast.rs
```diff
@@ -43,8 +43,8 @@ impl<T:Timestamp+Send> Progcaster<T> {
         let addr = path.clone();
         Progcaster {
             to_push: None,
-            pushers,
-            puller,
+            pushers: pushers.into_iter().map(|p| Box::new(p) as Box<dyn Push<_>>).collect(),
```
petrosagg, line 46: @frankmcsherry This is the only place I kept the trait objects because otherwise `Progcaster` would have to become generic over the pusher which would then add a generic parameter to `Subgraph`. I didn't know if that was ok but if it is I can eliminate this last bit of trait object if that sounds good

communication/src/allocator/thread.rs
```diff
@@ -26,11 +26,14 @@ pub struct Thread {
 }
 
 impl Allocate for Thread {
+    type Pusher<T: Data> = ThreadPusher<Message<T>>;
+    type Puller<T: Data> = ThreadPuller<Message<T>>;
+
     fn index(&self) -> usize { 0 }
     fn peers(&self) -> usize { 1 }
-    fn allocate<T: 'static>(&mut self, identifier: usize) -> (Vec<Box<dyn Push<Message<T>>>>, Box<dyn Pull<Message<T>>>) {
+    fn allocate<T: Data>(&mut self, identifier: usize) -> (Vec<Self::Pusher<T>>, Self::Puller<T>) {
```
antiguru, line 34: Possibly a breaking change as it requires all datums to be `Data`, which is currently not required on the same thread.

petrosagg: I don't know if this is the case because the trait itself defines this function as  `allocate<A: Data>`. I'm not sure why it was allowed to only specify `'static` here

https://docs.rs/timely_communication/0.12.0/timely_communication/allocator/trait.Allocate.html#tymethod.allocate

petrosagg: Looked a bit more into this and it's definitely not a breaking change. The `: Data` bound was being previously silently enforced so better to spell it out explicitly.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4a5b3884be978fa9b05a4eb5aabc1be8

petrosagg:
This PR now compiles on stable!

antiguru (approved):

PR #461, opened 2022-03-31
allow worker without dataflows to sleep

lluki:
As of now, a worker without any dataflows present will sit in a busy waiting loop. The thread will never park, as in `worker.rs` the thread parks only if the dataflow is not empty. 

This commit allows worker threads to `park` if no dataflow is installed. We unpark the thread if the `WorkerGuard` is dropped. But this introduces a race, as the unpark might be called before the thread is run. Thus, we introduce a per worker atomic bool `kill_switch` that gets set before we unpark the worker thread. The worker thread must now obey this flag: It must not park when the `kill_switch` is set.

This behavior is implemented in `worker.rs`. Clients of timely communication must obey this rule, and it breaks existing client code of communication, because the worker closure receives the `kill_switch` as additional argument.

### DIFF
--- a/communication/examples/comm_hello.rs
+++ b/communication/examples/comm_hello.rs
@@ -7,7 +7,7 @@ fn main() {
 
     // extract the configuration from user-supplied arguments, initialize the computation.
     let config = timely_communication::Config::from_args(std::env::args()).unwrap();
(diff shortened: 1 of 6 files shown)


frankmcsherry:
Superseded by #463. Thank you very much for finding, diagnosing, and fixing the issue, though! :D

PR #462, opened 2022-03-31
Specialize the container type for antichains

antiguru:
Antichains are currently unconditionally heap-allocated. For totally
ordered elements, an antichain will container zero or one elements, but not
more. This would allow us to avoid the heap allocation and instead store
the element inline.

To enable this, we add an associated type to `Timestamp` that defines a
container storing an antichain of the type. The container implements the
functionality to let the current `Antichain` type function without change.

The obvious downside of this change is that we now include a type bound
`T: Timestamp` on the `Antichain` type, because the elements are based on
`T`'s associated type, which infects a bunch of other places.


antiguru:
This PR has the deficit that it's not possible to create a `Antichain<Capability<T>>` anymore because it requires the generic parameter to `Antichain` to be `Timestamp`. `Capability` cannot implement `Timestamp` as it cannot offer a `minimum` function.

This functionality is required by Differential. I don't think there's a good work-around other than splitting the `Timestamp` trait in what it currently is plus a `TimstampStorage` that only carries the associated storage type, but it's rather ugly and I don't know that it actually works.

PR #465, opened 2022-04-27
fix a few broken links in docs

bddap:


frankmcsherry (approved):
Thanks for spotting these!

PR #467, opened 2022-05-24
nonce to prevent a split-brain communication cluster

antiguru:
Introduce a nonce value to ensure that all cluster members are connected to the same set of hosts.

### DIFF
--- a/communication/src/networking.rs
+++ b/communication/src/networking.rs
@@ -3,7 +3,7 @@
 use std::io;
 use std::io::{Read, Result};
 use std::net::{TcpListener, TcpStream};
-use std::sync::Arc;
+use std::sync::{Arc, Condvar, Mutex};
 use std::thread;
 use std::thread::sleep;
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


benesch:
> Is the PID sufficiently random? I just wanted to be sure that k8 or docker (or whatever runs in our cloud stack) doesn't always assigns the same PID...

Probably not! The PID is almost surely some stable small integer in Kubernetes. Generating a random nonce seems safer.

antiguru:
Changed it to use a random nonce.

Currently, we don't have a good way to report errors from `start_connections`. It might improve in the future.

communication/src/networking.rs
```diff
@@ -106,13 +127,19 @@ pub fn start_connections(addresses: Arc<Vec<String>>, my_index: usize, noisy: bo
 }
 
 /// Result contains connections [my_index + 1, addresses.len() - 1].
-pub fn await_connections(addresses: Arc<Vec<String>>, my_index: usize, noisy: bool) -> Result<Vec<Option<TcpStream>>> {
+pub fn await_connections(addresses: Arc<Vec<String>>, my_index: usize, noisy: bool, nonce: Arc<(Mutex<Option<u64>>, Condvar)>) -> Result<Vec<Option<TcpStream>>> {
     let mut results: Vec<_> = (0..(addresses.len() - my_index - 1)).map(|_| None).collect();
     let listener = TcpListener::bind(&addresses[my_index][..])?;
 
     for _ in (my_index + 1) .. addresses.len() {
         let mut stream = listener.accept()?.0;
```
umanwizard, line 135: This doesn't seem right. Since some processes might crash and then reconnect multiple times, we will end up with some of the connections being `None`.

antiguru:
Not needed anymore.

PR #468, opened 2022-06-08
store one element antichains inline

petrosagg:
Antichains are made to work with partially ordered times but that forces
them to allocate a heap vector even in simple cases of totally ordered
timestamp types which will never have more than one element in it.

Short of defining a `Frontier` trait and abstracting generically over
multiple independent frontier implementation this simple PR handles it
at runtime by backing `Antichain`s with `SmallVec`.

### DIFF
--- a/timely/src/progress/frontier.rs
+++ b/timely/src/progress/frontier.rs
@@ -1,5 +1,7 @@
 //! Tracks minimal sets of mutually incomparable elements of a partial order.
 
+use smallvec::{smallvec, SmallVec};
+
 use crate::progress::ChangeBatch;
 use crate::order::{PartialOrder, TotalOrder};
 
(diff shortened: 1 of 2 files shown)


timely/src/progress/frontier.rs
```diff
@@ -688,7 +690,7 @@ impl<'a, T: 'a> AntichainRef<'a, T> {
     ///```
     pub fn to_owned(&self) -> Antichain<T> where T: Clone {
         Antichain {
-            elements: self.frontier.to_vec()
+            elements: self.frontier.iter().cloned().collect()
```
aalexandrov, line 693: So you use `iter().cloned()` if you don't want to / can't consume an owned object and `into_iter()` if you can?

guswynn:
most timestamp times are gonna be <= the size of a usize, a vec is 3 usizes, and a smallvec appears to be the size of the array + 1 usize, so making the array's length smaller than 2 is probably just a waste!

antiguru (approved):
I prefer this PR over https://github.com/TimelyDataflow/timely-dataflow/pull/462 because it's less intrusive. #462 offers more opportunities for the compiler to optimize the generated code, but it comes at a usability burden.

I'm happy to approve the PR as-is. We can later change the size of the `SmallVec` backing storage, but for all total-order timestamps it's the right choice.

frankmcsherry:
This is present in `master` already at this point! Not sure how it got there, but both 1. thanks for the PR, and 2. closing nonetheless!

PR #471, opened 2022-07-15
Turn a stream of data into a stream of shared data

antiguru:
Add the `shared` function to streams, turning the contents of the stream
into `Rc` of the container. Downstream operators will see the same copy
of the data and the `Rc` takes care of cleaning up the shared reference.


antiguru:
I guess tests depend on #472 to pass!

frankmcsherry (approved):
Looks good also!

PR #472, opened 2022-07-15
Open the Inspect trait for more stream kinds

antiguru:
Instead of only supporting vector-backed streams, permit TimelyStack and
Rc streams.


frankmcsherry (approved):
Seems good to me!

PR #473, opened 2022-07-15
Extend RefOrMut to cover owned data.

antiguru:
The RefOrMut type previously covered immutable and mutable references to
data. We extend it to cover owned data in addition to the previous variants
to more clearly communicate that the sender doesn't want to hold on to the
data and also cannot accept an allocation instead.


antiguru:
Not needed anymore.

PR #475, opened 2022-07-15
Implement Filter for TimelyStack

antiguru:
As the title says, add an implementation for Filter for TimelyStack.

Note that the implementation does not free resources, it only removes elements from the container. The vector-based implementation does the same, but it can reclaim heap-allocated resources, which the `TimelyStack`-variant cannot.


antiguru:
Implemented as part of the core operators.

PR #484, opened 2022-10-04
Add missing .capture() to the first example

kishaningithub:


kishaningithub:
Can you kindly add "hacktoberfest-accepted" label to this PR / add "hacktoberfest" topic to this repository ?

More info - https://hacktoberfest.com/participation/#pr-mr-details

mdbook/src/chapter_0/chapter_0_0.md
```diff
@@ -10,7 +10,8 @@ use timely::dataflow::operators::{ToStream, Inspect};
 fn main() {
     timely::example(|scope| {
         (0..10).to_stream(scope)
-               .inspect(|x| println!("seen: {:?}", x));
+               .inspect(|x| println!("seen: {:?}", x))
+               .capture();
```
antiguru, line 14: It doesn't add any value to capture the output but then not use it. Capturing data is a mechanism to extract output from a dataflow. This example merely inspects the contents of a stream, but does not need the data outside the scope of Timely.

antiguru (changes requested):
This PR makes the initial example more complicated by adding a non-required function, without adding any explanation as to why this would be necessary.

mdbook/src/chapter_0/chapter_0_0.md
kishaningithub: But the given example does not compile in the latest version. I would personally expect the examples to not have any compilation error

antiguru: Not in front of a computer right now so I can't tell what the error might be. Please include the error in the description, and then we can determine the best fix.

antiguru: This PR doesn't build:
```
error[E0599]: no method named `capture` found for struct `StreamCore` in the current scope
   --> mdbook/src/chapter_0/chapter_0_0.md:14:17
    |
10  |                .capture();
    |                 ^^^^^^^ method not found in `StreamCore<timely::dataflow::scopes::Child<'_, Worker<timely::communication::allocator::Thread>, u64>, Vec<{integer}>>`
    |
   ::: /home/moritz/dev/repos/timely-dataflow/timely/src/dataflow/operators/capture/capture.rs:109:8
    |
109 |     fn capture(&self) -> ::std::sync::mpsc::Receiver<EventCore<T, D>> {
    |        ------- the method is available for `StreamCore<timely::dataflow::scopes::Child<'_, Worker<timely::communication::allocator::Thread>, u64>, Vec<{integer}>>` here
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
2   | use timely::dataflow::operators::Capture;
    |
```

kishaningithub: The original example builds correctly, the fault was mine, i seem to have missed a semicolon. This PR can be declined

PR #485, opened 2022-10-28
Use `Antichain` for `MutableAntichain::frontier`.

frankmcsherry:
This PR converts `MutableAntichain::frontier` from a `Vec<T>` to an `Antichain<T>`. In addition to a bit more code reuse, this allows `MutableAntichain` to expose a `&Antichain<T>` type rather than an `AntichainRef<T>`, which can be annoying to relate back to antichains (e.g. to `join` with, say). One can always go from an `&Antichain<T>` to an `AntichainRef<T>` with the `borrow()` method, but not the other way around.

The downside to scrutinize is the moment where we update `frontier` rebuilding the antichain. To my eyes, we do almost the same linear work per element, except that we additionally consider evicting elements that we know we wont have to evict. I'm ok losing that performance for the type clarity, or perhaps adding some `from_ordered_iter` to `Antichain`.

cc: @jkosh44

### DIFF
--- a/timely/src/progress/frontier.rs
+++ b/timely/src/progress/frontier.rs
@@ -293,6 +293,14 @@ impl<T> ::std::iter::IntoIterator for Antichain<T> {
     }
 }
 
+impl<'a, T> ::std::iter::IntoIterator for &'a Antichain<T> {
+    type Item = &'a T;
+    type IntoIter = ::std::slice::Iter<'a, T>;
+    fn into_iter(self) -> Self::IntoIter {
+        self.elements.iter()
+    }
+}
+
 /// An antichain based on a multiset whose elements frequencies can be updated.
 ///
 /// The `MutableAntichain` maintains frequencies for many elements of type `T`, and exposes the set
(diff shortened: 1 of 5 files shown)


jkosh44 (approved):
I'm just noticing this now, was I supposed to review this? If so, LGTM.

timely/src/progress/frontier.rs
```diff
@@ -557,16 +565,14 @@ impl<T> MutableAntichain<T> {
...
             self.changes.update(time, -1);
         }
 
         // build new frontier using strictly positive times.
         // as the times are sorted, we don't need to worry that we might displace frontier elements.
         for time in self.updates.iter().filter(|x| x.1 > 0) {
-            if !self.frontier.iter().any(|f| f.less_equal(&time.0)) {
-                self.frontier.push(time.0.clone());
-            }
+            self.frontier.insert(time.0.clone());
```
frankmcsherry, line 575: Consider `insert_ref`.

PR #486, opened 2022-11-04
Panic with consistent message on TCP failure

benesch:
To allow downstream crates to sniff out these panics and downgrade them, if appropriate.


philip-stoev:
Thanks a lot. From reading the code it seems that both TCP-related panics that I was after will be handled.

frankmcsherry:
Looks good to me, thanks!

PR #488, opened 2022-11-28
Activate operators that may want to shut down

frankmcsherry:
Operator shutdown was previously pretty loose, and only in response to operator activation. However, the conditions for shutdown can change without prompting an activation if e.g. a frontier becomes empty or a final capability is dropped. This meant that operators that should be shut down would instead linger until the dataflow itself is shut down. 

This PR adds that test as progress information is pushed to operators, in order to better clean up operators mid-dataflow. 

NB: Failing to shut down an operator should not have resulted in non-termination, unless operators were relying on dropping their state to signal something of consequence outward. All progress information would still be correct, and all downstream operators would receive correct frontiers.

### DIFF
--- a/timely/src/progress/subgraph.rs
+++ b/timely/src/progress/subgraph.rs
@@ -200,6 +200,7 @@ where
             incomplete_count,
             activations,
             temp_active: BinaryHeap::new(),
+            maybe_shutdown: Vec::new(),
             children: self.children,
             input_messages: self.input_messages,
             output_capabilities: self.output_capabilities,
@@ -242,6 +243,7 @@ where
     // shared activations (including children).
     activations: Rc<RefCell<Activations>>,
     temp_active: BinaryHeap<Reverse<usize>>,
+    maybe_shutdown: Vec<usize>,
 
     // shared state written to by the datapath, counting records entering this subgraph instance.
     input_messages: Vec<Rc<RefCell<ChangeBatch<TInner>>>>,
@@ -461,6 +463,7 @@ where
 
         // Drain propagated information into shared progress structure.
         for ((location, time), diff) in self.pointstamp_tracker.pushed().drain() {
+            self.maybe_shutdown.push(location.node);
             // Targets are actionable, sources are not.
             if let crate::progress::Port::Target(port) = location.port {
                 if self.children[location.node].notify {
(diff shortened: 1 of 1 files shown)


lluki:
Unfortunately it doesnt fix the operator leak of running TPC-H loadgen + materialized view with query 14. This is the situation after `drop materialized view q14;` and waiting 30s:

pr488

and this is `mz_dataflow_operator_dataflows`:
```
materialize=> set database to tpch;
SET
materialize=> drop materialized view q14;
DROP MATERIALIZED VIEW
materialize=> select * from mz_internal.mz_dataflow_operator_dataflows;
 id  |                   name                   | worker_id | dataflow_id |   dataflow_name   
-----+------------------------------------------+-----------+-------------+-------------------
 333 | Map                                      | 0         | 188         | Dataflow: 2.6.q14
 329 | FlatMap                                  | 0         | 188         | Dataflow: 2.6.q14
 331 | Exchange                                 | 0         | 188         | Dataflow: 2.6.q14
 326 | InspectBatch                             | 0         | 188         | Dataflow: 2.6.q14
 338 | InspectBatch                             | 0         | 188         | Dataflow: 2.6.q14
 345 | Dataflow: 2.6.q14                        | 0         | 188         | Dataflow: 2.6.q14
 188 | Dataflow: 2.6.q14                        | 0         | 188         | Dataflow: 2.6.q14
 328 | persist_sink u11 write_batches           | 0         | 188         | Dataflow: 2.6.q14
 340 | persist_sink u11 append_batches          | 0         | 188         | Dataflow: 2.6.q14
 323 | persist_sink u11 mint_batch_descriptions | 0         | 188         | Dataflow: 2.6.q14
(10 rows)

materialize=> show materialized views;
 name | cluster 
------+---------
(0 rows)

materialize=> 

```

frankmcsherry:
Ah, yes this wasn't meant to fix that for certain. It does fix a leak for e.g. `simple.rs`, and I'm happy to crack open the same example in MZ (the one based on `generate_series(0, large)`).

frankmcsherry:
Great, that appears to have the desired outcome, as we think that the remaining operators are the ones that haven't shut down, for whatever reason.

antiguru (approved):
This seems fine, as explained in the office hours!

One thing to keep in mind is that we could restrict the nodes added to `maybe_shutdown` to operators that have no inputs or the notify bit turned off.

PR #489, opened 2022-12-04
Make TCP communication fabric generic over network protocol

benesch:
This commit makes the communication fabric that was previously specific to TCP generic over the network protocol. Most practically this allows executing a timely computation over Unix domain sockets, which is useful in development, but in theory any network protocol that satisfies the new `Stream` trait could be used.


antiguru (approved):
This seems reasonable!

PR #494, opened 2023-01-11
offer mutable access to containers through handles

petrosagg:
This PR refactors a pattern that is pervasive to pretty much every single timely operator into an easier to use input handle method that also reduces the boilerplate code operator authors need to write.

The pattern is receiving a `RefOrMut` container from a handle and immediately swapping it with a local container in order to gain mutable access. In this repo all but one operators exhibited this pattern and associated boilerplate.


timely/src/dataflow/operators/generic/handles.rs
```diff
@@ -61,35 +59,43 @@ impl<'a, T: Timestamp, D: Container, P: Pull<BundleCore<T, D>>> InputHandleCore<
...
-    /// });
-    /// ```
+    /// Reads the next input buffer (at some timestamp `t`) and a corresponding capability for `t`.
+    /// The timestamp `t` of the input buffer can be retrieved by invoking `.time()` on the capability.
+    /// Returns `None` when there's no more data available.
     #[inline]
-    pub fn for_each<F: FnMut(CapabilityRef<T>, RefOrMut<D>)>(&mut self, mut logic: F) {
-        let mut logging = self.logging.take();
-        while let Some((cap, data)) = self.next() {
-            logging.as_mut().map(|l| l.log(crate::logging::GuardedMessageEvent { is_start: true }));
```
petrosagg, line 86: We can make the `next` and `next_mut` methods do this logging if desired. It wasn't clear to me if the current thinking was more in line with `next` that didn't log or with `for_each` that did log. Let me know if I got this the other way around

PR #495, opened 2023-01-15
update rustc in CI to 1.64

petrosagg:
Tests for the kafkaesque example started failing because it depends on `clap` which now requires rustc `1.64`. We could change the deps of the example but updating rustc seems like a good anything anyway.


frankmcsherry:
Thanks!

PR #496, opened 2023-01-18
Add cease to output handles

antiguru:
In certain situations, a handle survives longer than we would like to wait for a flush to happen. In this cases, an explicit call to cease can help to indicate to the rest of the system that no more data follows immediately, which is equivalent to dropping the handle.

Specifically, in async code the handle can be long-lived and survive await points, which makes it more important to signal momentary completion to the system.


frankmcsherry (approved):
Approving, though as discussed it's awkward that we can't really say what this function does other than flush some buffers. It does not seem to require (either idiomatically, or through the signature) that you actually "cease" so much as force a flush of the buffers.

PR #498, opened 2023-02-04
Turn RefOrMut of tuples to inner variants

antiguru:
Not sure how useful this is, but it would allow turning a RefOrMut of a tuple into a RefOrMut if its elements.


antiguru:
I haven't observed any case where this would have helped!

PR #499, opened 2023-02-09
Replace `finish` with `finish_non_exhaustive`

nooberfsh:


frankmcsherry:
This was folded up in #662, due to us having introduced a conflict with this PR. Seems like half of it was done earlier when a new argument was added, and the other half wasn't done at the same time. Thanks very much for the PR, though!

PR #501, opened 2023-02-15
Make the hello world example compile

def-:


frankmcsherry:
Hrm. This already passes for me at the moment:
```
mcsherry@dataflower timely-dataflow % cargo test
running 151 tests
...
test src/lib.rs - (line 25) ... ok
...
```
What are you running that it produces an error?

def-:
I see. I pasted it into a new file and ran it with `cargo run`.

frankmcsherry:
Ah, I see. `cargo test` will add some amount of goo to doctests such as linking in the crate under test and putting a `main` function around what you've written. Roughly to avoid having all the boiler plate show up for all the examples in the documentation.

PR #503, opened 2023-02-16
Fix ranges in chapter 2 of mdbook

def-:
Ranges with 0..9 have exclusive upper bounds in Rust, 0..=9 would be inclusive.

> The following program should print out the numbers one through ten.


frankmcsherry (approved):
Looks great to me!

PR #505, opened 2023-02-20
Back `MutableAntichain` by `ChangeBatch`

### LINKED ISSUE #504: Make MutableAntichain consolidate even without frontier advancement
Currently, `MutableAntichain` consolidates its contained `updates` only on frontier changes. That means when we keep pushing updates at a `MutableAntichain`, the size of that object keeps growing unless some of these updates also change the frontier.

For example, consider a `MutableAntichain` that tracks two frontiers, one that advances every second and one that doesn’t move. Assuming the moving frontier is beyond the non-moving frontier, the `MutableAntichain` receives one pair of (-1/+1) updates every second and keeps them all around forever.

To avoid memory leaks, `MutableAntichain` should regularly perform consolidation of its contents, even if its frontier does not advance. A simple strategy would be to consolidate each time the list of updates doubles in size, relative to the last time consolidation was performed.

frankmcsherry:
This PR removes the `Vec<(T, i64)>` from `MutableAntichain` and replaces it with a `ChangeBatch<T>`. There is a regression here in that several "unsafe" methods involving dirty bits are removed. Their justification was around progress tracking, and that justification no longer exists. This removes several ways in which one could write incorrect code (by dirtying things up and then reading data without first cleaning). These are now prevented.

Some potential for performance regressions in that certain loops are not exited as early as they are elsewhere, and the eager compaction work may not be helpful in all cases.

cc: @teskje

fixes #504

### DIFF
--- a/timely/src/progress/frontier.rs
+++ b/timely/src/progress/frontier.rs
@@ -308,14 +308,9 @@ impl<T> ::std::iter::IntoIterator for Antichain<T> {
 /// The `MutableAntichain` implementation is done with the intent that updates to it are done in batches,
 /// and it is acceptable to rebuild the frontier from scratch when a batch of updates change it. This means
 /// that it can be expensive to maintain a large number of counts and change few elements near the frontier.
-///
-/// There is an `update_dirty` method for single updates that leave the `MutableAntichain` in a dirty state,
-/// but I strongly recommend against using them unless you must (on part of timely progress tracking seems
-/// to be greatly simplified by access to this)
 #[derive(Clone, Debug, Abomonation, Serialize, Deserialize)]
 pub struct MutableAntichain<T> {
-    dirty: usize,
-    updates: Vec<(T, i64)>,
+    updates: ChangeBatch<T>,
     frontier: Vec<T>,
     changes: ChangeBatch<T>,
 }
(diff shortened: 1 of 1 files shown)


benesch (approved):
Well, makes sense to me, for whatever that's worth! Seems like a good thing to slip into Materialize at the start of a release cycle (e.g., now), so we can get some nightly coverage with the new implementation.

timely/src/progress/frontier.rs
```diff
@@ -502,32 +469,20 @@ impl<T> MutableAntichain<T> {
...
+                rebuild_required = !(beyond_frontier || (delta < 0 && before_frontier));
+            }
 
-            self.dirty -= 1;
+            self.updates.update(time, delta);
         }
-        self.dirty = 0;
 
         if rebuild_required {
             self.rebuild()
```
antiguru, line 488: Nit below: Adjust `rebuild` comment?

timely/src/progress/frontier.rs
```diff
@@ -396,13 +379,12 @@ impl<T> MutableAntichain<T> {
     /// assert!(frontier.frontier() == AntichainRef::new(&[0u64]));
     ///```
     #[inline]
-    pub fn new_bottom(bottom: T) -> MutableAntichain<T>
-    where
-        T: Clone,
+    pub fn new_bottom(bottom: T) -> MutableAntichain<T> 
+    where
```
antiguru, line 383: Spaces?

antiguru (approved):
LGTM!

PR #507, opened 2023-02-23
Log messages over enter/leave channels

teskje:
This PR adds `MessageEvent` logging for messages traveling over `enter`/`leave` channels.

The motivation is to fix missing record counts in MZ's memory graphs: https://github.com/MaterializeInc/materialize/issues/14443.

### DIFF
--- a/timely/src/dataflow/operators/enterleave.rs
+++ b/timely/src/dataflow/operators/enterleave.rs
@@ -21,6 +21,7 @@
 
 use std::marker::PhantomData;
 
+use crate::logging::{TimelyLogger, MessagesEvent};
 use crate::progress::Timestamp;
 use crate::progress::timestamp::Refines;
 use crate::progress::{Source, Target};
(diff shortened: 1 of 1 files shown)


teskje:
Some evidence that his actually helps:

* With the current timely main:
Screenshot 2023-02-23 at 15 24 38

* With this branch:
Screenshot 2023-02-23 at 15 16 50

timely/src/dataflow/operators/enterleave.rs
```diff
@@ -90,17 +91,22 @@ impl<G: Scope, T: Timestamp+Refines<G::Timestamp>, C: Data+Container> Enter<G, T
...
             active: false,
         };
         let produced = ingress.targets.produced().clone();
-
         let input = scope.subgraph.borrow_mut().new_input(produced);
-
         let channel_id = scope.clone().new_identifier();
-        self.connect_to(input, ingress, channel_id);
-        StreamCore::new(Source::new(0, input.port), registrar, scope.clone())
+        let pusher = LogPusher::new(ingress, channel_id, scope);
```
antiguru, line 101: We could skip creating `pusher` if `scope.logging()` is None. That's not what we do elsewhere, but maybe we should?

timely/src/dataflow/operators/enterleave.rs
```diff
@@ -130,8 +136,11 @@ impl<'a, G: Scope, D: Clone+Container, T: Timestamp+Refines<G::Timestamp>> Leave
 
         let output = scope.subgraph.borrow_mut().new_output();
         let (targets, registrar) = TeeCore::<G::Timestamp, D>::new();
+        let egress = EgressNub { targets, phantom: PhantomData };
         let channel_id = scope.clone().new_identifier();
-        self.connect_to(Target::new(0, output.port), EgressNub { targets, phantom: PhantomData }, channel_id);
+        let pusher = LogPusher::new(egress, channel_id, &scope);
```
antiguru, line 141: Same here.

antiguru (approved):

timely/src/dataflow/operators/enterleave.rs
teskje: Is there a way to do that efficiently? The only I can come up with is using a `Box<dyn Push>`, and I'm not clear about the performance implications of that. Does it trade of the `if let Som(logger)` check with a vtable lookup on each push?

```diff
  let produced = ingress.targets.produced().clone();
  let input = scope.subgraph.borrow_mut().new_input(produced);
  let channel_id = scope.clone().new_identifier();
- let pusher = LogPusher::new(ingress, channel_id, scope);
+ let pusher: Box<dyn Push<_>> = match scope.logging() {
+     Some(logger) => Box::new(LogPusher::new(ingress, channel_id, scope.index(), logger)),
+     None => Box::new(ingress),
+ };

  self.connect_to(input, pusher, channel_id);
```

teskje: As discussed, the trick to making this efficient is to duplicate the `connect_to` call.

PR #514, opened 2023-03-12
Inline functions frequently called from Materialize

antiguru:
Observed from applying callgrind on Materialize.

### DIFF
--- a/container/src/columnation.rs
+++ b/container/src/columnation.rs
@@ -23,6 +23,7 @@ impl<T: Columnation> TimelyStack<T> {
     ///
     /// Note that the associated region is not initialized to a specific capacity
     /// because we can't generally know how much space would be required.
+    #[inline(always)]
     pub fn with_capacity(capacity: usize) -> Self {
         Self {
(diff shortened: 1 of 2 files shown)


container/src/columnation.rs
```diff
@@ -81,6 +84,7 @@ impl<T: Columnation> TimelyStack<T> {
     /// Retain elements that pass a predicate, from a specified offset.
     ///
     /// This method may or may not reclaim memory in the inner region.
+    #[inline]
```
antiguru, line 87: Check if this is actually helpful.

timely/src/progress/frontier.rs
```diff
@@ -495,6 +495,7 @@ impl<T> MutableAntichain<T> {
     /// This method is meant to be used for bulk updates to the frontier, and does more work than one might do
     /// for single updates, but is meant to be an efficient way to process multiple updates together. This is
     /// especially true when we want to apply very large numbers of updates.
+    #[inline]
```
antiguru, line 498: Remove.

PR #515, opened 2023-03-17
Remove core

### LINKED ISSUE #510: Remove `Core` variants of various types
We added several `Core` variants of types to support different container types. For example, `Stream<G, D>` is really a `StreamCore<G, Vec<D>>`. We should remove the `Core` variants and make them the default instead.

This is a breaking change, but it seems we have sufficient experience with the container abstraction that we can make the container type with its contents to the default stream element.

The following types need to be adjusted:
 * [ ] `StreamCore`
 * [ ] `TeeCore`
 * [ ] `EventProducerCore`
 * [ ] `EventConsumerCore`
 * [ ] `BundleCore`
 * [ ] `ParallelizationContractCore`
 * [ ] `ExchangeCore`
 * [ ] `BufferCore`
 * [ ] `AutoflushSessionCore`
 * [ ] `CounterCore`
 * [ ] `feedback::HandleCore`
 * [ ] `input::HandleCore`
 * [ ] `UnorderedHandleCore`
 * [ ] `EventIteratorCore`
 * [ ] `EventPusherCore`
 * [ ] `EventLinkCore`
 * [ ] `EventWriterCore`
 * [ ] `EventReaderCore`
 * [ ] `InputHandleCore`
 * [ ] `FrontierdInputHandleCore`
 * [ ] `OutputHandleCore`

antiguru:
Warning: breaking changes up ahead!

Fixes #510 

This change removes the `Core` variants of types, traits, and functions that we introduced in #426. However, this means it introduces breaking changes, specifically where we used `Data` parameters in the past we now use `Container`. This is desirable because maintaining the old facade comes with a cognitive overhead. After this change, most places that previously specified individual elements now specify containers. All places that only apply to specific (mostly `Vec<_>`) containers are exepted.

The most notable breaking changes are in inputs, `to_stream`, and feedback, which now need to know the kind of container they produce. In the past, they were fixed to `Vec<_>`, but now the caller needs to be explicit about it.

The PR looks rather large, but most of the changes are purely mechanical. For wildly-used types, there is a commit that inlines the non-core variant, and a follow-up to rename the core-variant to the old name. This should make it explicit that each change has limited scope. Tests should pass after each individual change.
* 60ae900977ae8e1f03098f60a4696343cd77c0bb changes inputs to containers.
* 5987d4cfd4bab462a101f790093f5d468d752783 changes feedback to use containers.

### DIFF
--- a/kafkaesque/src/bin/capture_recv.rs
+++ b/kafkaesque/src/bin/capture_recv.rs
@@ -3,7 +3,6 @@ use timely::dataflow::operators::capture::Replay;
 use timely::dataflow::operators::Accumulate;
 
 use rdkafka::config::ClientConfig;
-
 use kafkaesque::EventConsumer;
 
 fn main() {
@@ -31,7 +30,7 @@ fn main() {
             .filter(|i| i % worker.peers() == worker.index())
             .map(|i| {
                 let topic = format!("{}-{:?}", topic, i);
-                EventConsumer::<_,u64>::new(consumer_config.clone(), topic)
+                EventConsumer::<_, Vec<u64>>::new(consumer_config.clone(), topic)
             })
             .collect::<Vec<_>>();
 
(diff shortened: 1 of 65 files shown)


antiguru:
TimelyDataflow/differential-dataflow#383 shows what changes we'd need to make in Differential to switch to the new interface.

antiguru:
Superseded by various other changes.

PR #522, opened 2023-05-26
book: Correct mismatched Stateful Operator references to `maximum` from `max`.

arusahni:
At some point this example fell out of sync with the body.


antiguru (approved):
Thank you, this looks good!

PR #525, opened 2023-05-30
input handles: give empty containers to operators

### LINKED ISSUE #523: Passing empty containers to `Session::give_container` can stall downstream frontier
It appears that passing an empty container to `Session::give_container` will sometimes stall the frontier of downstream operators, i.e. prevent them from advancing even though they should.

This is the minimal example I could come up with:
```rust
extern crate timely;

use timely::dataflow::channels::pact::Pipeline;
use timely::dataflow::operators::{Exchange as _, Input, InspectCore, Operator, Probe};
use timely::dataflow::InputHandle;

fn main() {
    timely::execute_from_args(std::env::args(), |worker| {
        let mut buf = Vec::new();
        let probe = worker.dataflow::<u64, _, _>(|scope| {
            scope
                .input_from(&mut input)
                .unary(Pipeline, "Test", move |_, _| {
                    move |input, output| {
                        input.for_each(|cap, data| {
                            data.swap(&mut buf);
                            let mut session = output.session(&cap);
                            session.give_container(&mut Vec::new());
                            session.give_container(&mut buf);
                        });
                    }
                })
                .exchange(|x| *x)
                .probe()
        });

        for round in 0..2 {
            input.send(round);
            input.advance_to(round + 1);
        }
        input.close();

        while !probe.done() {
            worker.step();
        }

        println!("worker {} complete", worker.index());
    })
    .unwrap();
}
```

We would expect this example to eventually print "worker <n> complete". Instead it just hangs.

Some initial observations:
* I tried adding `inspect_container` after the various operators, to print frontier advancements.
    * When adding it after the `unary` operator, the hang does not occur.
    * When adding it after the `exchange` operator, it only prints frontier `[1]`.
* When any of the two `give_container` calls is removed, the hang does not occur.

petrosagg:
The current code tried to be helpful by ignoring any messages with zero length, but that made it impossible for the caller to differentiate between "There was an empty container message" vs "There are no more messages".

In the presence of empty containers in the channels this can lead operators to prematurely yield back to timely without fully processing their inputs, which in turn can lead to the whole dataflow not making progress since timely won't re-activate operators that have leftover events in their channels.

This PR removes this check and always forwards the containers to the operators, exactly as they were sent.

Fixes #523

### DIFF
--- a/timely/src/dataflow/channels/pullers/counter.rs
+++ b/timely/src/dataflow/channels/pullers/counter.rs
@@ -46,15 +46,12 @@ impl<T:Ord+Clone+'static, D: Container, P: Pull<BundleCore<T, D>>> Counter<T, D,
     #[inline]
     pub(crate) fn next_guarded(&mut self) -> Option<(ConsumedGuard<T>, &mut BundleCore<T, D>)> {
         if let Some(message) = self.pullable.pull() {
-            if message.data.len() > 0 {
-                let guard = ConsumedGuard {
-                    consumed: Rc::clone(&self.consumed),
-                    time: Some(message.time.clone()),
-                    len: message.data.len(),
-                };
-                Some((guard, message))
-            }
-            else { None }
+            let guard = ConsumedGuard {
+                consumed: Rc::clone(&self.consumed),
+                time: Some(message.time.clone()),
+                len: message.data.len(),
+            };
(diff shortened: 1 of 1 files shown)


antiguru (approved):
LGTM, thank you!

PR #526, opened 2023-06-04
Activate only by channel ID

antiguru:
This changes activation to avoid an associated event, which is currently not used anymore. It also switches to a `Vec` to contain the activations, and deduplicates before activating operators.

This improves performance in edge cases:
```
➜  timely-dataflow git:(activation_no_event) time taskset 1 target/release/examples/pingpong $((2**20)) 64
taskset 1 target/release/examples/pingpong $((2**20)) 64  3.63s user 0.00s system 99% cpu 3.638 total
➜  timely-dataflow-master git:(master) time taskset 1 target/release/examples/pingpong $((2**20)) 64
taskset 1 target/release/examples/pingpong $((2**20)) 64  3.32s user 0.78s system 99% cpu 4.106 total
```

### DIFF
--- a/communication/src/allocator/counters.rs
+++ b/communication/src/allocator/counters.rs
@@ -2,7 +2,6 @@
 
 use std::rc::Rc;
 use std::cell::RefCell;
-use std::collections::VecDeque;
 
 use crate::{Push, Pull};
 use crate::allocator::Event;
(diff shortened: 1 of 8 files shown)


communication/src/allocator/counters.rs
```diff
@@ -2,7 +2,6 @@
 
 use std::rc::Rc;
 use std::cell::RefCell;
-use std::collections::VecDeque;
 
 use crate::{Push, Pull};
 use crate::allocator::Event;
```
antiguru, line 7: Check if `Event` still needs to exist.

timely/src/worker.rs
```diff
@@ -338,7 +338,9 @@ impl<A: Allocate> Worker<A> {
             let events = allocator.events().clone();
             let mut borrow = events.borrow_mut();
             let paths = self.paths.borrow();
-            for (channel, _event) in borrow.drain(..) {
+            borrow.sort_unstable();
+            borrow.dedup();
+            for channel in borrow.drain(..) {
                 // TODO: Pay more attent to `_event`.
```
antiguru, line 344: Update comment.

frankmcsherry (approved):
This makes sense to me, and we've left the comment in that reminds us how it might be different in the future (tracking the emptiness of channels, and not scheduling only on the basis of a message both sent and received).

PR #533, opened 2023-10-13
Remove worker identifier from logging

antiguru:
This changes the data received from logging from (Duration, Id, Data) to
(Duration, Data), saving the space for Id. In most cases, the Id can be
derived from the context where the log data is processed.

### DIFF
--- a/communication/src/logging.rs
+++ b/communication/src/logging.rs
@@ -20,6 +20,8 @@ pub enum CommunicationEvent {
     Message(MessageEvent),
     /// A state transition.
     State(StateEvent),
+    /// Setup event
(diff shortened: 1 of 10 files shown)


communication/src/logging.rs
```diff
@@ -20,6 +20,8 @@ pub enum CommunicationEvent {
     Message(MessageEvent),
     /// A state transition.
     State(StateEvent),
+    /// Setup event
+    Setup(CommunicationSetup)
```
frankmcsherry, line 24: Are we certain this actually gets constructed anywhere? I didn't see it, but I may have missed it.

PR #534, opened 2023-10-13
Remove an unused dependency on clap

antiguru:
This prevents building Timely in CI because recent versions of clap require a newer version of Rust than what we use in CI.


frankmcsherry:
I mean, seems good. No clue how this happened, but I bet we used it in examples or somesuch! :D

PR #535, opened 2023-11-01
Implement Columnation for Product

antiguru:
Implement `Columnation` for Product, which it was missing.


frankmcsherry (approved):
Looks good from here.

PR #536, opened 2023-11-25
Insert an element by reference into an antichain

antiguru:
This is useful if all the caller has is a reference.


frankmcsherry (approved):
Looks great; thank you very much!

PR #537, opened 2023-11-26
Manual Antichain::default to avoid bounds

antiguru:
The derived `default` for Antichain includes a bound `T: Default`, which is unnecessary. This change replaces the derived implementation with a custom one that does not have the constrain.


frankmcsherry:
Looks good; thanks!

PR #538, opened 2023-11-26
DNM: Remove Container::Item

antiguru:
It's not used, and unclear as to what it should communicate. Let's talk this over!

### DIFF
--- a/timely/src/dataflow/operators/inspect.rs
+++ b/timely/src/dataflow/operators/inspect.rs
@@ -9,7 +9,7 @@ use crate::dataflow::{Scope, StreamCore};
 use crate::dataflow::operators::generic::Operator;
 
 /// Methods to inspect records and batches of records on a stream.
-pub trait Inspect<G: Scope, C: Container>: InspectCore<G, C> + Sized {
+pub trait Inspect<G: Scope, C: Container, I>: InspectCore<G, C> + Sized {
     /// Runs a supplied closure on each observed data element.
     ///
     /// # Examples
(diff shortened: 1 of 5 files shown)


timely/src/dataflow/operators/reclock.rs
```diff
@@ -45,11 +45,11 @@ pub trait Reclock<S: Scope> {
     /// assert_eq!(extracted[1], (5, vec![4,5]));
     /// assert_eq!(extracted[2], (8, vec![6,7,8]));
     /// ```
-    fn reclock<TC: Container<Item=()>>(&self, clock: &StreamCore<S, TC>) -> Self;
```
antiguru, line 48: This is only looking for a progress stream, but requires a container because we have no other way to express its expectations.

timely/src/dataflow/operators/inspect.rs
```diff
@@ -84,25 +84,25 @@ pub trait Inspect<G: Scope, C: Container>: InspectCore<G, C> + Sized {
...
 }
 
-impl<G: Scope, D: Data+Columnation> Inspect<G, TimelyStack<D>> for StreamCore<G, TimelyStack<D>> {
+impl<G: Scope, D: Data+Columnation> Inspect<G, TimelyStack<D>, D> for StreamCore<G, TimelyStack<D>> {
     fn inspect_core<F>(&self, mut func: F) -> Self where F: FnMut(Result<(&G::Timestamp, &[D]), &[G::Timestamp]>) + 'static {
         self.inspect_container(move |r| func(r.map(|(t, c)| (t, &c[..]))))
     }
 }
 
-impl<G: Scope, C: Container> Inspect<G, Rc<C>> for StreamCore<G, Rc<C>>
```
antiguru, line 102: This changes the semantics for inspecting `Rc<C>`, and I'm not sure we use this. It might be relevant for consuming streams of differential updates at some point, but at the moment it's `Vec<Rc<Batch>>` and not `Rc<Container<Batch>>` for some unspecified container.

PR #540, opened 2023-11-29
Remove Container: Clone + 'static

antiguru:
Remove the requirement that all Container implementations are Clone and 'static. This makes implementing types simpler that depend on Container, and requires us to explicitly mark various places as `Data` such that they comply with Timely's type requirements.

### DIFF
--- a/container/src/columnation.rs
+++ b/container/src/columnation.rs
@@ -339,7 +339,7 @@ mod container {
 
     use crate::columnation::{Columnation, TimelyStack};
 
-    impl<T: Columnation + 'static> Container for TimelyStack<T> {
+    impl<T: Columnation> Container for TimelyStack<T> {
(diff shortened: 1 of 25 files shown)


frankmcsherry:
I started to look at this, and one of the first complaints from rust about removing `Clone` is the number of `RefOrMut::swap` calls we have. These are potentially an antipattern with containers, in that we'll likely be expecting to read a bunch of references out rather than owned contents, and cloning the container (and contents) while close to mandatory for `Ref` variants normally, probably isn't what we want with containers.

Probably not a big deal at the moment in that I think most `Ref` variants are e.g. `abomonation` related, and that's a different antipattern. But we might want to have a think about the right way to encourage good behavior here, and whether we are even doing the right thing ourselves.

PR #541, opened 2024-02-04
Container GATs, improve traits

antiguru:
Cleanup the `Container::Item` type, move it to `PushPartitioned` and make it a GAT. Similarly, split the `Map` trait in two (`Map`, `MapInPlace`) and support GATs in the `Map` trait.

This has the benefit that Timely becomes less opinionated about the interface provided by containers, for example by allowing non-owned data when calling `stream.map`.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-22.04
     steps:
       - uses: actions/checkout@v3
-      - run: cargo install mdbook --version 0.4.27
+      - run: cargo install mdbook --version 0.4.36
       - run: cd mdbook && mdbook build
       - uses: JamesIves/github-pages-deploy-action@v4
         with:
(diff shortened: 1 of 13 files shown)


antiguru:
I reworked the PR to only contain a minimum of changes. This compiles with differential once applying the following patch:

```diff
diff --git a/src/collection.rs b/src/collection.rs
index f2daf8ad..20cab0d2 100644
--- a/src/collection.rs
+++ b/src/collection.rs
@@ -416,10 +416,10 @@ impl<G: Scope, D: Data, R: Semigroup> Collection<G, D, R> where G::Timestamp: Da
     ///          .inspect_batch(|t,xs| println!("errors @ {:?}: {:?}", t, xs));
     /// });
     /// ```
-    pub fn inspect_batch<F>(&self, func: F) -> Collection<G, D, R>
+    pub fn inspect_batch<F>(&self, mut func: F) -> Collection<G, D, R>
     where F: FnMut(&G::Timestamp, &[(D, G::Timestamp, R)])+'static {
         self.inner
-            .inspect_batch(func)
+            .inspect_batch(move |time, batch| func(time, &batch[..]))
             .as_collection()
     }
     /// Attaches a timely dataflow probe to the output of a Collection.
(
```

frankmcsherry (approved):
This looks good! I propose we merge this and start to see what it leads to downstream.

PR #546, opened 2024-02-05
Revert "async: relax requirements of supplied streams (#401)"

antiguru:
This reverts commit c5195eb and 935b7e9.


frankmcsherry (approved):
This looks great, and lines up with the goal of reducing TD down to its essential complexity.

PR #562, opened 2024-04-28
Container builder

antiguru:
Introduce a `ContainerBuilder` that sits on the outputs of operators and gets a chance to process data as produced by the operator. While the PR breaks many APIs, it tries to preserve compatibility for all users of `output.session`, which we hard code to a builder that's based on the length and capacity of a container. In other places, we need a type annotation, specifically when there is no call to `session`, or just a call to `session_with_builder`, which is the equivalent to `session` but doesn't force a specific builder.

The PR also replaces `AutoflushSession` with `AutoflushSessionCore`, and renames the latter to avoid `Core`. All uses were internal and easy to fix, which should apply to external crates, too.

This is compatible with differential to offer in-place consolidation.

Do not look commit-by-commit as the history represents a worklog instead of a grouping of changes. Readers are encouraged to first look at the `timely-container` library.

### DIFF
--- a/container/src/columnation.rs
+++ b/container/src/columnation.rs
@@ -310,7 +310,7 @@ mod serde {
 
 mod container {
     use std::ops::Deref;
-    use crate::{Container, ContainerBuilder, PushContainer, PushInto};
+    use crate::{Container, PushContainer};
 
     use crate::columnation::{Columnation, TimelyStack};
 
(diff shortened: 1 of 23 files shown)


antiguru:
https://github.com/TimelyDataflow/differential-dataflow/pull/478 shows how to integrate this with differential.

antiguru:
Oh, if we want to have the return-position impl trait, we'd need to move to Rust 1.75. This, or refactoring to use explicit types both work for me.

frankmcsherry (approved):
Looks good, thank you!

PR #563, opened 2024-05-06
Move OutputHandleCore::cease to more general impl

antiguru:
Move `OutputHandleCore::cease` to a more general implementation. This fixes a leftover from the previous implementation that was not aware of container builders.


frankmcsherry (approved):
lgtm!

PR #564, opened 2024-05-20
Complete PushInto implementations

antiguru:
This adds some PushInto implementations we forgot about.

### DIFF
--- a/container/src/columnation.rs
+++ b/container/src/columnation.rs
@@ -4,6 +4,8 @@ use std::iter::FromIterator;
 
 pub use columnation::*;
 
+use crate::PushInto;
+
 /// An append-only vector that store records as columns.
 ///
 /// This container maintains elements that might conventionally own
@@ -235,6 +237,27 @@ impl<T: Columnation> Clone for TimelyStack<T> {
         for item in &source[..] {
             self.copy(item);
         }
+    }
+}
+
+impl<T: Columnation> PushInto<TimelyStack<T>> for T {
+    #[inline]
+    fn push_into(self, target: &mut TimelyStack<T>) {
+        target.copy(&self);
+    }
+}
+
+impl<T: Columnation> PushInto<TimelyStack<T>> for &T {
+    #[inline]
+    fn push_into(self, target: &mut TimelyStack<T>) {
+        target.copy(self);
+    }
+}
+
+impl<T: Columnation> PushInto<TimelyStack<T>> for &&T {
+    #[inline]
+    fn push_into(self, target: &mut TimelyStack<T>) {
+        target.copy(self);
     }
 }
 
--- a/container/src/lib.rs
+++ b/container/src/lib.rs
@@ -233,10 +233,17 @@ impl<T> PushInto<Vec<T>> for T {
     }
 }
 
-impl<'a, T: Clone> PushInto<Vec<T>> for &'a T {
+impl<T: Clone> PushInto<Vec<T>> for &T {
     #[inline]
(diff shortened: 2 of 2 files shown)


container/src/columnation.rs
```diff
@@ -238,6 +240,27 @@ impl<T: Columnation> Clone for TimelyStack<T> {
...
+        target.copy(self);
+    }
+}
+
+impl<T: Columnation> PushInto<TimelyStack<T>> for &&T {
+    #[inline]
+    fn push_into(self, target: &mut TimelyStack<T>) {
+        target.copy(self);
+    }
+}
```
frankmcsherry, line 262: Maybe silly, but can this be `impl<T: PushInto<TimelyStack<T>> .. for &T`? Would that get both of these, and also `&&&T`?

frankmcsherry (approved):
Makes sense. I'm not sure I understand the need for the `&&T` implementations, though. Should those be discouraged and the use changed to `&T`, or is there a reason not to do `&&&T` also? I left a note about a possible blanket implementation, but if it doesn't work no worries.

container/src/columnation.rs
antiguru: Yeah, this is a curiosity! We currently don't need in timely, but we might need it at some point. It arises when reading out of a container that returns references and copying into a new container. So while the items itself might be a reference, the reading can add another reference, which requires adding the `&&` implementation. Concretely, consider a container for slices, and we'd like to copy `[T]` in it. This means that the receiver needs to absorb `&T`, because that's the only thing we can get out of a slice. Now, we want `T` to be a container itself, and suddently we have nested references. This line in flatcontainer requires it:  (( You wouldn't be wrong to say it's quite esoteric, but, hey, it works :) ))
https://github.com/antiguru/flatcontainer/blob/main/src/lib.rs#L670

antiguru:
Blanket for all types does not work, unsurprisingly:
```
error[E0275]: overflow evaluating the requirement `TimelyStack<&&&&&&...>: Sized`
   |
   = help: consider increasing the recursion limit by adding a `#![recursion_limit = "256"]` attribute to your crate (`timely_container`)
note: required for `&'a &_` to implement `for<'a> PushInto<TimelyStack<&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&_>>`
  --> container/src/lib.rs:66:12
   |
66 | impl<T, C> PushInto<C> for &&T where for<'a> &'a T: PushInto<C> {
   |         -  ^^^^^^^^^^^     ^^^
   |         |
   |         unsatisfied trait bound introduced here
   = note: 125 redundant requirements hidden
   = note: required for `&'a &&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&_` to implement `for<'a> PushInto<TimelyStack<&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&&_>>`
```

PR #566, opened 2024-05-24
PushInto targets container instead of item

antiguru:
Update to flatcontainer 0.3.0, which changes the `CopyInto` trait into a `Push` trait. Update Timely's `PushInto` trait in the same manner, i.e., the parameter is what should be pushed and self is the container.

Split `PushContainer` into a `push` function on `Container` and a `SizableContainer` that knows `capacity`, `preferred_capacity`, and `resize` functions, which the capacity container builder and `PushPartitioned` require. It also appears as a restriction on `ContainerBuilder`, as did `PushContainer` in the past.

Adds a implementation for `flatcontainer::Region` for `TimelyStack`, it's not used yet but would allow us to store columnated data in flatcontainer.

### DIFF
--- a/container/src/lib.rs
+++ b/container/src/lib.rs
@@ -246,6 +246,13 @@ impl<T: Clone> PushInto<&T> for Vec<T> {
     }
 }
 
+impl<T: Clone> PushInto<&&T> for Vec<T> {
+    #[inline]
+    fn push_into(&mut self, item: &&T) {
+        self.push_into(*item)
+    }
+}
+
 mod rc {
     use std::ops::Deref;
     use std::rc::Rc;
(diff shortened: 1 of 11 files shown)


container/src/lib.rs
```diff
@@ -226,24 +231,18 @@ impl<T: Clone + 'static> PushContainer for Vec<T> {
...
 
-impl<T: Clone> PushInto<Vec<T>> for &&T {
+impl<T: Clone> PushInto<&T> for Vec<T> {
     #[inline]
-    fn push_into(self, target: &mut Vec<T>) {
-        (*self).push_into(target);
+    fn push_into(&mut self, item: &T) {
+        self.push(item.clone())
     }
 }
```
frankmcsherry, line 247: It looks like you lost the `PushInto<&&T>` implementation. If it's not needed, no worries, but wanted to check!

frankmcsherry (approved):
Looks good. A few questions, but nothing deeply consequential.

PR #568, opened 2024-06-03
Add reborrow functions to Container

antiguru:
Add `reborrow` and `reborrow_ref` functions to `Container` to shorten the lifetime of the GATs.


antiguru:
Turns out this might actually not be needed for what I had in mind. Might still be useful, tho?

PR #569, opened 2024-06-07
Rework container builder to use push into

antiguru:
Previously, `ContainerBuilder` had `push` and `push_container` functions,
which were odd in the presence of the `PushInto` trait. This change removes
the functions and instead relies on a `PushInto` implementation. As a
bonus, this removes several `SizableContainer` bounds, which are now
up to the caller to enforce should they push into a capacity-based
container builder.

Specifically, it adds the following new types or APIs:
* ContainerBuilder: remove `push` and `push_container`. Replaces the fromer
  with a `PushInto` implementation, and moves the latter into a function
  on `CapacityContainerBuilder`. All uses of `give_container` are now
  special-cased to said builder. Other builders can accept containers
  through their `PushInto` implementation.
* A `BufferingContainerBuilder` that only accepts complete buffers. Could
  back out that change because it's similar (but not equal!) to the
  capacity-based variant.
* core::Input learns `new_input_with_buider` that allows the user to
  specify a customer builder. Existing APIs should function unchanged and
  use the capacity-based builder.
* core::SharedStream uses the buffering container builder.
* core::to_stream gains a `ToStreamBuilder` trait that allows the user to
  specify the builder. `ToStream` uses that but fixes the builder to the
  capacity-based variant.

### DIFF
--- a/container/src/lib.rs
+++ b/container/src/lib.rs
@@ -93,7 +93,7 @@ pub trait PushInto<T> {
 /// needs to produce all outputs, even partial ones. Caller should repeatedly call the functions
 /// to drain pending or finished data.
 ///
-/// The caller is responsible to fully consume the containers returned by [`Self::extract`] and
+/// The caller should consume the containers returned by [`Self::extract`] and
 /// [`Self::finish`]. Implementations can recycle buffers, but should ensure that they clear
 /// any remaining elements.
 ///
@@ -109,9 +109,11 @@ pub trait ContainerBuilder: Default + 'static {
     /// be called repeatedly, for example while the caller can send data.
     ///
     /// Returns a `Some` if there is data ready to be shipped, and `None` otherwise.
+    #[must_use]
     fn extract(&mut self) -> Option<&mut Self::Container>;
     /// Extract assembled containers and any unfinished data. Should
     /// be called repeatedly until it returns `None`.
+    #[must_use]
     fn finish(&mut self) -> Option<&mut Self::Container>;
 }
 
@@ -132,6 +134,7 @@ pub struct CapacityContainerBuilder<C>{
 }
 
 impl<T, C: SizableContainer + PushInto<T>> PushInto<T> for CapacityContainerBuilder<C> {
+    #[inline]
     fn push_into(&mut self, item: T) {
         if self.current.capacity() == 0 {
             self.current = self.empty.take().unwrap_or_default();
--- a/container/src/lib.rs
+++ b/container/src/lib.rs
@@ -23,7 +23,7 @@ pub trait Container: Default + Clone + 'static {
     /// The type of elements when reading non-destructively from the container.
     type ItemRef<'a> where Self: 'a;
 
-    /// The type of elements when draining the continer.
+    /// The type of elements when draining the container.
     type Item<'a> where Self: 'a;
 
     /// Push `item` into self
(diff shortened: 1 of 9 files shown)


timely/src/dataflow/operators/core/input.rs
```diff
@@ -274,26 +357,48 @@ impl<T: Timestamp, C: Container> Handle<T, C> {
...
-                debug_assert!(self.buffer2.is_empty());
+        while let Some(container) = self.builder.finish() {
+            Self::flush_container(container, &mut self.buffer, &mut self.pushers, &self.now_at);
+        }
+    }
+
+    /// flushes our buffer at each of the destinations. there can be more than one; clone if needed.
+    /// Does not take `self` because `flush` and `extract` borrow `self` mutably.
+    #[inline(never)]
+    fn flush_container(
```
frankmcsherry, line 379: Future TODO: We should rationalize the words used here to e.g. perhaps change this to `push_container` or .. eh .. `extract_container` or something that communicates "move some amount of work out of the handle" more than "be done". Also fix up `close_epoch` similarly.

timely/src/dataflow/operators/core/input.rs
```diff
@@ -274,26 +357,48 @@ impl<T: Timestamp, C: Container> Handle<T, C> {
...
+        container: &mut CB::Container,
+        buffer: &mut CB::Container,
+        pushers: &mut [Counter<T, CB::Container, Tee<T, CB::Container>>],
+        now_at: &T
+    ) {
+        for index in 0 .. pushers.len() {
+            if index < pushers.len() - 1 {
+                buffer.clone_from(container);
+                Message::push_at(buffer, now_at.clone(), &mut pushers[index]);
+                debug_assert!(buffer.is_empty());
```
frankmcsherry, line 389: Seems maybe stronger than strictly needed, as `buffer` could be holding resources that might be useful. Perhaps a container should be able to hold resources even across a `clear()` call, and the assert can pass, but e.g. a `Vec<String>` might not want to be cleared, so that `clone_from(container)` can use the string allocations themselves. Or not, maybe too complicated.

timely/src/dataflow/operators/core/rc.rs
```diff
@@ -27,13 +28,13 @@ pub trait SharedStream<S: Scope, C: Container> {
 impl<S: Scope, C: Container> SharedStream<S, C> for StreamCore<S, C> {
     fn shared(&self) -> StreamCore<S, Rc<C>> {
         let mut container = Default::default();
-        self.unary(Pipeline, "Shared", move |_, _| {
+        self.unary::<BufferingContainerBuilder<_>, _, _, _>(Pipeline, "Shared", move |_, _| {
```
frankmcsherry, line 31: We discussed, and it seems possible that this could be a `CapacityContainerBuilder` and we largely use the old code, using `give_container` and prompting a pushthrough.

timely/src/dataflow/operators/core/to_stream.rs
```diff
@@ -55,3 +60,31 @@ impl<C: SizableContainer, I: IntoIterator+'static> ToStream<C> for I where C: Pu
         })
     }
 }
+
+/// Converts to a timely [StreamCore]. Equivalent to [`ToStreamBuilder`] but
+/// uses a [`CapacityContainerBuilder`].
+pub trait ToStream<C: Container> {
```
frankmcsherry, line 66: It feels weird that we allow folks to have opinions about the containers used without having opinions about how those containers are populated. Like "use a `Vec` please" but no opinions on what the vec should contain.

frankmcsherry (approved):
Looks great, with some comments about potentially optional code, and some clarity / moments where I was confused as a reader.

container/src/lib.rs
```diff
@@ -170,26 +188,58 @@ impl<C: Container> ContainerBuilder for CapacityContainerBuilder<C> {
             self.pending.push_back(std::mem::replace(container, empty));
         }
     }
+}
+
+/// A container builder that absorbs entire containers. Maintains FIFO order.
```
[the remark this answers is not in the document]
antiguru: Removed.

frankmcsherry (approved):
Looks good. Merge at your discretion!

PR #571, opened 2024-06-14
Clone for Config and other improvements

antiguru:
* Clone for timely and communication config. Forces the cluster log function to be wrapped in an Arc, which is what happens anyways, so a no-op change.
* Use `execute_from` in `execute` to deduplicate code.
* Remove weird logging code from `execute`.

### DIFF
--- a/communication/src/allocator/zero_copy/initialize.rs
+++ b/communication/src/allocator/zero_copy/initialize.rs
@@ -39,7 +39,7 @@ pub fn initialize_networking(
     my_index: usize,
     threads: usize,
     noisy: bool,
-    log_sender: Arc<Box<dyn Fn(CommunicationSetup)->Option<Logger<CommunicationEvent, CommunicationSetup>>+Send+Sync>>,
+    log_sender: Arc<dyn Fn(CommunicationSetup)->Option<Logger<CommunicationEvent, CommunicationSetup>>+Send+Sync>,
 )
 -> ::std::io::Result<(Vec<TcpBuilder<ProcessBuilder>>, CommsGuard)>
 {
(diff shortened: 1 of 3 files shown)


communication/src/allocator/zero_copy/initialize.rs
```diff
@@ -39,7 +39,8 @@ pub fn initialize_networking(
     my_index: usize,
     threads: usize,
     noisy: bool,
-    log_sender: Box<dyn Fn(CommunicationSetup)->Option<Logger<CommunicationEvent, CommunicationSetup>>+Send+Sync>)
+    log_sender: Arc<Box<dyn Fn(CommunicationSetup)->Option<Logger<CommunicationEvent, CommunicationSetup>>+Send+Sync>>,
```
frankmcsherry, line 42: I think you can skip the `Box`, as `Arc<dyn Trait>` is a thing!

timely/src/execute.rs
```diff
@@ -218,6 +219,17 @@ where
 /// // the extracted data should have data (0..10) thrice at timestamp 0.
 /// assert_eq!(recv.extract()[0].1, (0..30).map(|x| x / 3).collect::<Vec<_>>());
 /// ```
+///
+/// ### Communication logging
+///
+/// For multi-process (cluster) configurations, this functions installs a custom log
+/// function, replacing any previous function. The function connects to the host in
+/// the environment symbol `TIMELY_COMM_LOG_ADDR`.
```
frankmcsherry, line 227: Should we just delete this feature?

frankmcsherry (approved):

communication/src/allocator/zero_copy/initialize.rs
antiguru: Nice!

timely/src/execute.rs
antiguru: Tossed!

PR #575, opened 2024-08-20
Remove `abomonation` to reduce unsoundness

frankmcsherry:
The `unsafe` keyword is the source of unsoundness in Rust, and we would like to minimize it. One particular villain (by its own admission) is abomonation, which provides unsafe serialization and deserialization in the name of performance. Although performance is great, the unsoundness that results from using `unsafe` is not.

This PR does not evaluate any performance considerations, nor does it even test whether `bincode` is a suitable replacement in the contexts in which it is used. Probably worth shaking that out before merging anything.

### DIFF
--- a/communication/src/allocator/zero_copy/push_pull.rs
+++ b/communication/src/allocator/zero_copy/push_pull.rs
@@ -90,7 +90,7 @@ impl<T:Data> Pull<Message<T>> for Puller<T> {
         self.receiver
             .borrow_mut()
             .pop_front()
-            .map(|bytes| Message::from_bytes(bytes));
+            .map(Message::from_bytes);
 
         &mut self.current
     }
@@ -134,7 +134,7 @@ impl<T:Data> Pull<Message<T>> for PullerInner<T> {
             self.receiver
                 .borrow_mut()
                 .pop_front()
-                .map(|bytes| Message::from_bytes(bytes));
+                .map(Message::from_bytes);
 
             &mut self.current
         }
--- a/communication/src/allocator/zero_copy/push_pull.rs
+++ b/communication/src/allocator/zero_copy/push_pull.rs
@@ -90,7 +90,7 @@ impl<T:Data> Pull<Message<T>> for Puller<T> {
         self.receiver
             .borrow_mut()
             .pop_front()
-            .map(|bytes| unsafe { Message::from_bytes(bytes) });
+            .map(|bytes| Message::from_bytes(bytes));
 
         &mut self.current
     }
@@ -134,7 +134,7 @@ impl<T:Data> Pull<Message<T>> for PullerInner<T> {
             self.receiver
                 .borrow_mut()
                 .pop_front()
-                .map(|bytes| unsafe { Message::from_bytes(bytes) });
+                .map(|bytes| Message::from_bytes(bytes));
 
             &mut self.current
         }
--- a/communication/src/logging.rs
+++ b/communication/src/logging.rs
@@ -1,7 +1,9 @@
 //! Configuration and events for communication logging.
 
+use serde::{Serialize, Deserialize};
+
 /// Configuration information about a communication thread.
-#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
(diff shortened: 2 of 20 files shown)


communication/src/allocator/zero_copy/push_pull.rs
```diff
@@ -90,7 +90,7 @@ impl<T:Data> Pull<Message<T>> for Puller<T> {
         self.receiver
             .borrow_mut()
             .pop_front()
-            .map(|bytes| unsafe { Message::from_bytes(bytes) });
+            .map(|bytes| Message::from_bytes(bytes));
```
antiguru, line 93:
```suggestion
            .map(Message::from_bytes);
```

communication/src/allocator/zero_copy/push_pull.rs
```diff
@@ -134,7 +134,7 @@ impl<T:Data> Pull<Message<T>> for PullerInner<T> {
             self.receiver
                 .borrow_mut()
                 .pop_front()
-                .map(|bytes| unsafe { Message::from_bytes(bytes) });
+                .map(|bytes| Message::from_bytes(bytes));
```
antiguru, line 137:
```suggestion
                .map(Message::from_bytes);
```

communication/src/logging.rs
```diff
@@ -1,7 +1,7 @@
 //! Configuration and events for communication logging.
 
 /// Configuration information about a communication thread.
-#[derive(Abomonation, Debug, PartialEq, Eq, Hash, Clone, Copy)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
```
antiguru, line 4: Shouldn't we derive `Serialize`/`Deserialize` here?

antiguru (approved):
Looks good! There are some logging types that implemented abomination, but don't support serde. Should they? We might not cover their construction and serialization in any of the tests.

communication/src/logging.rs
frankmcsherry: Good point. I didn't check that any more than `cargo test` passed, so it's very possible much has been missed. It looks like just a scan for "removed `Abomination`, but doesn't have serde" would be valuable. I'll do that!

communication/src/allocator/zero_copy/push_pull.rs
frankmcsherry: Actually, .. I'm not even sure that `zero_copy/` makes any sense without Abomination (certainly `bincode` is not zero copy, so .. does the concept even make sense?{.

communication/src/networking.rs
```diff
@@ -21,7 +21,7 @@ type ByteOrder = byteorder::BigEndian;
 /// Framing data for each `Vec<u8>` transmission, indicating a typed channel, the source and
 /// destination workers, and the length in bytes.
 // *Warning*: Adding, removing and altering fields requires to adjust the implementation below!
-#[derive(Abomonation, Debug, PartialEq, Eq, Hash, Clone, Copy)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
```
frankmcsherry, line 24: Adding in `, Serialize, Deserialize` here (one of the logging events has one of these, and can't derive without it). That might change how we want to do the reading and writing below, but happy to fix that up later.

communication/src/allocator/zero_copy/push_pull.rs
frankmcsherry: Tidied, and note left.

communication/src/networking.rs
frankmcsherry: Added back in.

frankmcsherry:
Rebased, and I *think* that all types that previously derived `Abomonation` now derive both `Serialize` and `Deserialize`. In the future there could be different traits to use (e.g. `Columnable`, or some other way of communicating "there is a SOA representation that can be `memcpy`'d around"), but .. I think we are good-ish to go?

antiguru (approved):

frankmcsherry:
Thanks for the review!

PR #581, opened 2024-09-06
Introduce `SmallVec` for small allocations

frankmcsherry:
This PR introduce the use of `SmallVec` for allocations backing `ChangeBatch` and `Antichain`. These often have at most two and one, respectively, elements in them. I made `ChangeBatch` itself generic, and `Antichain` not. Probably best to unify these. 

It takes a modified `event_driven.rs` from 
```
cargo run --example event_driven --release -- 1000 1000 no  5.31s user 0.59s system 96% cpu 6.124 total
```
to
```
cargo run --example event_driven --release -- 1000 1000 no  4.56s user 0.59s system 95% cpu 5.382 total
```
Some noise is these measurements, but the relative improvement seems consistent.

### DIFF
--- a/timely/Cargo.toml
+++ b/timely/Cargo.toml
@@ -31,6 +31,7 @@ timely_logging = { path = "../logging", version = "0.12" }
 timely_communication = { path = "../communication", version = "0.12", default-features = false }
 timely_container = { path = "../container", version = "0.12" }
 crossbeam-channel = "0.5.0"
+smallvec = { version = "1.13.2", features = ["serde", "const_generics"] }
 
 [dev-dependencies]
 # timely_sort="0.1.6"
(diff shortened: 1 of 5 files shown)


antiguru (approved):
Looks good, thanks!

timely/examples/event_driven.rs
```diff
@@ -34,7 +34,7 @@ fn main() {
 
         println!("{:?}\tdataflows built ({} x {})", timer.elapsed(), dataflows, length);
 
-        for round in 0 .. {
+        for round in 0 .. 10 {
```
antiguru, line 37: We might not want to ship this change tho.

frankmcsherry: oops.

PR #582, opened 2024-09-06
Shared reference-counted operator path

antiguru:
Encapsulate the path for an operator as a `Rc<[usize]>` that can be shared
among all interested parties, instead of passing owned vectors around. This
avoids some memory allocations for rendering dataflow graphs.

### DIFF
--- a/timely/src/dataflow/scopes/child.rs
+++ b/timely/src/dataflow/scopes/child.rs
@@ -127,7 +127,7 @@ where
         let index = self.subgraph.borrow_mut().allocate_child_id();
         let path = self.addr_for_child(index);
 
-        let subscope = RefCell::new(SubgraphBuilder::new_from(index, path, self.logging(), self.progress_logging.clone(), name));
+        let subscope = RefCell::new(SubgraphBuilder::new_from(path, self.logging(), self.progress_logging.clone(), name));
         let result = {
             let mut builder = Child {
                 subgraph: &subscope,
--- a/timely/src/dataflow/scopes/child.rs
+++ b/timely/src/dataflow/scopes/child.rs
@@ -58,10 +58,10 @@ where
     fn config(&self) -> &Config { self.parent.config() }
     fn index(&self) -> usize { self.parent.index() }
     fn peers(&self) -> usize { self.parent.peers() }
-    fn allocate<D: Data>(&mut self, identifier: usize, address: Vec<usize>) -> (Vec<Box<dyn Push<Message<D>>>>, Box<dyn Pull<Message<D>>>) {
+    fn allocate<D: Data>(&mut self, identifier: usize, address: Rc<[usize]>) -> (Vec<Box<dyn Push<Message<D>>>>, Box<dyn Pull<Message<D>>>) {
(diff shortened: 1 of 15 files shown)


timely/src/dataflow/scopes/child.rs
```diff
@@ -125,7 +125,7 @@ where
         F: FnOnce(&mut Child<Self, T2>) -> R,
     {
         let index = self.subgraph.borrow_mut().allocate_child_id();
-        let path = self.subgraph.borrow().path.clone();
+        let path = self.addr_for_child(index);
```
frankmcsherry, line 128: I can believe this change is correct, but I don't understand it.

antiguru: It's hidden in the change to `SubgraphBuilder::new_from`, which doesn't add the index to the path anymore. Maybe it's better to not pass `index` but provide a complete path to said function?

frankmcsherry: I see, so with the change below, rather than pass in `index` and `path`, and have the method append one into the other, we do it ahead of time here with `addr_for_child` and `new_from` just takes the path but uses it in a different sense (previously: "path to `self`" and now: "path to the subscope")?

antiguru: Yup, that's the idea. We could go back the the former implementation, but then might need another allocation for the dataflow root or so.

timely/src/progress/subgraph.rs
```diff
@@ -94,20 +94,19 @@ where
         self.edge_stash.push((source, target));
     }
 
-    /// Creates a new Subgraph from a channel allocator and "descriptive" indices.
+    /// Creates a new Subgraph from a channel allocator and "descriptive" indices. The Subgraph's
+    /// index is the last element of `path`.
     pub fn new_from(
```
frankmcsherry, line 99: I think this is better now! At least, I can't really tell why it made sense to provide the index and the path separately, and there is no `self` involved for which the `path` would make sense. I'm tempted to remove the line in the comment that you added, because "The Subgraph's index" is no longer a concept that anyone needs to know about. I would be tempted to "say more" in the comments, but the previous comments were so vague that it's hard to require it. In fact, .. I think the comments are just wrong? There is no channel allocator here yet.

What do you think about 
```rust
/// Creates a `SubgraphBuilder` from a path of indexes from the dataflow root to the subgraph, terminating with the local index of the new subgraph itself.
```

timely/src/dataflow/scopes/child.rs
frankmcsherry: No, this looks better, and having read the call site for `new_from`, it just seems strictly better this way (clearer, even). Just, the process of realizing this was more complicated than the other changes from `Vec<T>` to `Rc<[T]>`. :D

frankmcsherry (approved):
Ok, I totally get all the `Vec<T>` into `Rc<[T]>` stuff and think that is great. I was surprised by some other changes, but I do now think they are both correct and strict improvements in clarity, to boot! Sorry for the hold-up, and thank you for the explanations!

PR #587, opened 2024-09-29
ci: Update to `actions/checkout@v4`

waywardmonkeys:
This just updates to the current version which runs on a version of Node that GitHub hasn't deprecated.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,7 +9,7 @@ jobs:
   deploy:
     runs-on: ubuntu-22.04
     steps:
-      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This might fail due to getting an updated version of `unicode-width` due to an MSRV issue where the latest version now needs 1.66 but this is building with 1.65.

That's unrelated to the changes I did here.

frankmcsherry:
Thanks! I'll need to sync with @antiguru as I have less of a clue wrt how all these bits operate. My guess is it's all good, and shouldn't need more than a few business hours minutes to confirm (and find a new MSRV, to get things green again).

antiguru (approved):
Thanks, looks good!

PR #589, opened 2024-09-30
Update CI, bump minimum Rust to 1.78

antiguru:
Update the CI tooling with what Differential uses. Bump Rust to 1.78, which is the lowest version supported by Differential.


frankmcsherry (approved):
LGTM!

PR #590, opened 2024-09-30
Improve docs.

waywardmonkeys:
This mainly adds missing backticks and fixes some unmatched backticks.

PR #591, opened 2024-10-01
Use `dep` syntax with `getopts`

waywardmonkeys:
By using `dep:getopts`, there isn't a need to rename the package to `getopts_dep`.


waywardmonkeys:
This will conflict with #590, but I will happily rebase either this or that over top of whichever lands first.

frankmcsherry:
Happy to merge, but also happy to let the rebase shake out. IIRC, the optional dependency only exists because some folks wanted to try and use DD as part of Rust's build chain, and you weren't allowed to depend on `getopts` there. But .. tbh it's probably ok to rip out at this point (they haven't been back in touch).

PR #593, opened 2024-10-16
Allow viewing of next dataflow-unique identifier

frankmcsherry:
This PR adds a simple method that allows one to view the next dataflow-unique identifier that will be allocated. The intended use is for folks who want to correlate spans in rendering code with spans of constructed operators and channels. One can view the identifier before and after a block of rendering code, and the resulting span of identifiers should align with the logged identifiers for created operators and channels.

@mgree


antiguru (approved):
Looks good!

PR #596, opened 2024-11-10
Simplify communication `Message` type

frankmcsherry:
This PR chases down some simplifications due to the removal of `abomonation`, looking forward to flexibility enabled by containers.

Specifically, the `Message<T>` type has wrapped several ways to present a `T`, including owned and borrowed forms, as well as numerous accessors. All of the borrowed forms have been dead code since the `abomonation` removal, and the attendant complexity unwarranted. This PR removes it.

Going forward, distinctions between owned and borrowed seem best housed in containers, which can themselves resemble `MessageContents<T>` which was the enum that could be owned or borrowed. The `Message<T>` type remains, but seemingly only to support serialization methods, but this behavior could be move into a trait that containers are expected to implement if they are to be used for exchange edges.

### DIFF
--- a/communication/src/message.rs
+++ b/communication/src/message.rs
@@ -5,8 +5,6 @@ use crate::Data;
 
 /// Either an immutable or mutable reference.
 pub enum RefOrMut<'a, T> where T: 'a {
-    /// An immutable reference.
-    Ref(&'a T),
     /// A mutable reference.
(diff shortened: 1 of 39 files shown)


antiguru (approved):
Looks good. Didn't read in detail as I'm on the phone.

PR #597, opened 2024-11-10
Move opinions about encoding from `communication` to `timely`.

frankmcsherry:
This PR introduces a trait `Bytesable` with methods from and into `Bytes`. It removes from `communication` any reliance on `bincode`, not because it is bad just because it is an opinion. It almost removes all dependence on `serde`, except that it creates several logging types that want to derive `Serialize` and `Deserialize`, but it is not otherwise reliant on them.

The trait that was formerly `communication::Data` is now instead `communication::allocator::Exchangeable`, and it looks like so:
```rust
/// A type that can be sent along an allocated channel.
pub trait Exchangeable : Send+Any+Bytesable+'static { }
impl<T: Send+Any+Bytesable+'static> Exchangeable for T { }
```

In exchange, timely gets a relatively small opinion, currently in the `encoding` module, that one can implement `Bytesable` with some reliance on `serde` and `bincode`. The opinions are expressed in a trait
```rust
    /// A composite trait for types that may be used with channels.
    pub trait Data : Send+Any+Serialize+for<'a>Deserialize<'a>+'static { }
    impl<T: Send+Any+Serialize+for<'a>Deserialize<'a>+'static> Data for T { }
```
which informs the `timely::ExchangeData` trait.

### DIFF
--- a/communication/examples/comm_hello.rs
+++ b/communication/examples/comm_hello.rs
@@ -1,5 +1,24 @@
-use std::ops::Deref;
-use timely_communication::{Message, Allocate};
+use timely_communication::{Allocate, Bytesable};
+
+/// A wrapper that indicates `bincode` as the serialization/deserialization strategy.
+pub struct Message {
+    /// Text contents.
+    pub payload: String,
+}
+
+impl Bytesable for Message {
+    fn from_bytes(bytes: timely_bytes::arc::Bytes) -> Self {
+        Message { payload: std::str::from_utf8(&bytes[..]).unwrap().to_string() }
+    }
+
+    fn length_in_bytes(&self) -> usize {
+        self.payload.len()
+    }
+
+    fn into_bytes<W: ::std::io::Write>(&self, writer: &mut W) {
+        writer.write_all(self.payload.as_bytes()).unwrap();
+    }
+}
 
 fn main() {
 
@@ -14,7 +33,7 @@ fn main() {
 
         // send typed data along each channel
         for i in 0 .. allocator.peers() {
-            senders[i].send(Message::from_typed(format!("hello, {}", i)));
+            senders[i].send(Message { payload: format!("hello, {}", i)});
             senders[i].done();
         }
 
@@ -26,7 +45,7 @@ fn main() {
             allocator.receive();
 
             if let Some(message) = receiver.recv() {
-                println!("worker {}: received: <{}>", allocator.index(), message.deref());
+                println!("worker {}: received: <{}>", allocator.index(), message.payload);
                 received += 1;
             }
 
--- a/communication/examples/comm_hello.rs
+++ b/communication/examples/comm_hello.rs
@@ -1,6 +1,6 @@
 use timely_communication::{Allocate, Bytesable};
 
-/// A wrapper that indicates `bincode` as the serialization/deserialization strategy.
+/// A wrapper that indicates the serialization/deserialization strategy.
 pub struct Message {
     /// Text contents.
     pub payload: String,
--- a/communication/src/allocator/mod.rs
+++ b/communication/src/allocator/mod.rs
@@ -85,8 +85,8 @@ pub trait Allocate {
     /// By default, this method uses the thread-local channel constructor
     /// based on a shared `VecDeque` which updates the event queue.
     fn pipeline<T: 'static>(&mut self, identifier: usize) ->
-        (thread::ThreadPusher<Message<T>>,
-         thread::ThreadPuller<Message<T>>)
+        (thread::ThreadPusher<T>,
+         thread::ThreadPuller<T>)
     {
         thread::Thread::new_from(identifier, self.events().clone())
     }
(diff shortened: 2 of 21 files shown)


frankmcsherry:
One thing I'd should tidy up are the unit tests in the doccomments. I copy/pasted in something that I knew would work, but it is a departure from what was there before. I don't know that they are bad, just .. different unintentionally.

frankmcsherry:
`9d2ac9e` moves the opinions all the way to the pacts, which are where we construct the channels. The default pacts use the `bincode` serialization strategy, but .. in the worst case the pacts are pluggable and can be implemented for other strategies. But probably the next thing to think through is how to have an opinion on the pact without having an opinion on the serialization.

communication/examples/comm_hello.rs
```diff
@@ -1,7 +1,26 @@
 extern crate timely_communication;
 
-use std::ops::Deref;
-use timely_communication::{Message, Allocate};
+use timely_communication::{Allocate, Bytesable};
+
+/// A wrapper that indicates `bincode` as the serialization/deserialization strategy.
```
antiguru, line 5:
```suggestion
/// A wrapper that supports serializing strings to bytes.
```

communication/src/allocator/mod.rs
```diff
@@ -32,6 +32,12 @@ pub trait AllocateBuilder : Send {
     fn build(self) -> Self::Allocator;
 }
 
+use std::any::Any;
+
+/// A type that can be sent along an allocated channel.
+pub trait Exchangeable : Send+Any+Bytesable+'static { }
+impl<T: Send+Any+Bytesable+'static> Exchangeable for T { }
```
antiguru, line 39:
```suggestion
pub trait Exchangeable : Send+Any+Bytesable { }
impl<T: Send+Any+Bytesable> Exchangeable for T { }
```

`Any` implies `'static'`.

communication/src/initialize.rs
```diff
@@ -182,47 +182,71 @@ impl Config {
...
-///         allocator.receive();
-///         if let Some(message) = receiver.recv() {
-///             use std::ops::Deref;
-///             println!("worker {}: received: <{}>", allocator.index(), message.deref());
-///             expecting -= 1;
-///         }
-///         allocator.release();
+/// use timely_communication::{Allocate, Bytesable};
+/// 
+/// /// A wrapper that indicates `bincode` as the serialization/deserialization strategy.
```
antiguru, line 187: Fix comment.

timely/src/lib.rs
```diff
@@ -114,5 +114,55 @@ impl<T: Clone+'static> Data for T { }
...
+mod encoding {
+
+    use std::any::Any;
+    use serde::{Serialize, Deserialize};
+    use timely_bytes::arc::Bytes;
+    use timely_communication::Bytesable;
+
+    /// A composite trait for types that may be used with channels.
+    pub trait Data : Send+Any+Serialize+for<'a>Deserialize<'a>+'static { }
+    impl<T: Send+Any+Serialize+for<'a>Deserialize<'a>+'static> Data for T { }
```
antiguru, line 132:
```suggestion
    pub trait Data : Send+Any+Serialize+for<'a>Deserialize<'a> { }
    impl<T: Send+Any+Serialize+for<'a>Deserialize<'a>> Data for T { }
```

communication/src/allocator/mod.rs
```diff
@@ -32,6 +32,12 @@ pub trait AllocateBuilder : Send {
     fn build(self) -> Self::Allocator;
 }
 
+use std::any::Any;
+
+/// A type that can be sent along an allocated channel.
+pub trait Exchangeable : Send+Any+Bytesable+'static { }
```
antiguru, line 38: Maybe move to crate's root?

communication/src/allocator/process.rs
```diff
@@ -110,7 +110,7 @@ impl Process {
 impl Allocate for Process {
     fn index(&self) -> usize { self.index }
     fn peers(&self) -> usize { self.peers }
-    fn allocate<T: Any+Send+Sync+'static>(&mut self, identifier: usize) -> (Vec<Box<dyn Push<Message<T>>>>, Box<dyn Pull<Message<T>>>) {
+    fn allocate<T: Any+Send+'static>(&mut self, identifier: usize) -> (Vec<Box<dyn Push<T>>>, Box<dyn Pull<T>>) {
```
antiguru, line 113:
```suggestion
    fn allocate<T: Any+Send>(&mut self, identifier: usize) -> (Vec<Box<dyn Push<T>>>, Box<dyn Pull<T>>) {
```

antiguru (approved):
Looks good, I think this is a worthwhile simplification. Some minor suggestions around leaving out `'static` when it's implied by `Any`. No strong opinion though, it mostly removes some visual clutter.

communication/src/allocator/mod.rs
frankmcsherry: Could re-publish there, but I think all of the uses are in this module (and in other crates).

frankmcsherry:
I'll take the `communication` documentation simplification as a future action item, to keep things moving along.

PR #598, opened 2024-11-11
Rust updates, better doc testing

antiguru:
Capture a bunch of changes I had sitting around for a while. Fixes a bug in the mdbook, but otherwise no functional changes.

Some of the changes might conflict with #597.


frankmcsherry:
Looks good. I'll handle fix-ups for #597.

PR #610, opened 2024-12-17
dataflow/operators/core: add `.partition()` for `StreamCore`

petrosagg:


timely/src/dataflow/operators/core/partition.rs
```diff
@@ -0,0 +1,73 @@
...
+    ///                          .partition(3, |x| (x % 3, x));
+    ///
+    ///     for (idx, stream) in streams.into_iter().enumerate() {
+    ///         stream
+    ///             .container::<Vec<_>>()
+    ///             .inspect(move |x| println!("seen {idx}: {x:?}"));
+    ///     }
+    /// });
+    /// ```
+    fn partition<C2, D2, F>(&self, parts: u64, route: F) -> Vec<StreamCore<G, C2>>
```
antiguru, line 30: Consider taking a container builder here instead!

petrosagg: Good point! Changed to a container builder argument

frankmcsherry:
Looks good to me, thanks!

timely/src/dataflow/operators/core/partition.rs
```diff
@@ -0,0 +1,75 @@
...
+    ///         stream
+    ///             .container::<Vec<_>>()
+    ///             .inspect(move |x| println!("seen {idx}: {x:?}"));
+    ///     }
+    /// });
+    /// ```
+    fn partition<CB, D2, F>(&self, parts: u64, route: F) -> Vec<StreamCore<G, CB::Container>>
+    where
+        CB: ContainerBuilder,
+        CB::Container: SizableContainer + PushInto<D2> + Data,
```
antiguru, line 33: The `SizableContainer` requirement here is not needed, right? That's a requirement that'll be imposed by the container builder, but the implementation should only require the remaining traits for the container.

antiguru: Fixed in #612.

PR #614, opened 2024-12-27
Align `Bytesable` messages to `u64`

frankmcsherry:
The `Bytesable` trait is what indicates how we want to serialize messages, should we need to do that. There is an advantage to data containers being `u64` aligned, specifically for `columnar` implementations but potentially for other implementations. This PR adds up to 7 bytes for each message to ensure that each part ends up `u64` aligned. With these changes, the `columnar` example does not need to relocate (memcpy) in order to get aligned data.

Tbd: other containers (e.g. `Vec`, `FlatStack`) don't have the same defensive measures in place yet, and we'll want to figure out how to prevent them from messing anything up either. Their `ContainerBytes` implementation could do this, or the `Bytesable` implementation for `Message` could round up once it has serialized everything.

### DIFF
--- a/timely/src/dataflow/channels/mod.rs
+++ b/timely/src/dataflow/channels/mod.rs
@@ -74,7 +74,6 @@ where
         let time_size = ::bincode::serialized_size(&time).expect("bincode::serialized_size() failed") as usize;
         // We expect to find the `data` payload at `8 + 8 + round_up(time_size)`;
         let bytes_read = 8 + 8 + ((time_size + 7) & !7);
-        // let bytes_read = bytes.len() - slice.len();
         bytes.extract_to(bytes_read);
         let data: C = ContainerBytes::from_bytes(bytes);
         Self { time, data, from, seq }
@@ -113,6 +112,8 @@ pub trait ContainerBytes {
 
 mod implementations {
 
+    use std::io::Write;
+
     use serde::{Serialize, Deserialize};
     use crate::dataflow::channels::ContainerBytes;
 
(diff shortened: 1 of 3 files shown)


timely/src/lib.rs
```diff
@@ -108,8 +108,8 @@ impl<T: Clone+'static> Data for T { }
 ///
 /// The `ExchangeData` trait extends `Data` with any requirements imposed by the `timely_communication`
 /// `Data` trait, which describes requirements for communication along channels.
-pub trait ExchangeData: Data + encoding::Data { }
-impl<T: Data + encoding::Data> ExchangeData for T { }
+pub trait ExchangeData: Data + encoding::Data + columnar::Columnar { }
+impl<T: Data + encoding::Data + columnar::Columnar> ExchangeData for T { }
```
frankmcsherry, line 112: This .. worked out ok but I realized is from a prior attempt that wanted to columnarize things. Unnecessary for this PR.

antiguru: I think we should remove it from this PR because it'll require us to change Differential.

timely/src/dataflow/channels/mod.rs
```diff
@@ -57,7 +57,9 @@ impl<T, C: Container> Message<T, C> {
 }
 
 // Instructions for serialization of `Message`.
-// Intended to swap out the constraint on `C` for `C: Bytesable`.
+//
+// Serialization of each field is meant to be `u64` aligned, so that each has tha ability
```
antiguru, line 61:
```suggestion
// Serialization of each field is meant to be `u64` aligned, so that each has the ability
```

antiguru (approved):
LGTM modulus comments

PR #615, opened 2025-01-08
Define loggers in terms of container builders

antiguru:
The logging infrastructure had some old assumptions built in, such as the
container type and the size of buffers. With this change, we defer to the
container builder pattern to re-use the existing infrastructure. This also
allows us to switch the container type to something else if we'd like to do
so.

### DIFF
--- a/communication/src/allocator/zero_copy/tcp.rs
+++ b/communication/src/allocator/zero_copy/tcp.rs
@@ -2,14 +2,16 @@
 
 use std::io::{self, Write};
 use crossbeam_channel::{Sender, Receiver};
-use timely_logging::Logger;
-use crate::logging::CommunicationEventBuilder;
-use crate::logging::{CommunicationEvent, MessageEvent, StateEvent};
+
 use crate::networking::MessageHeader;
 
 use super::bytes_slab::BytesSlab;
 use super::bytes_exchange::MergeQueue;
 use super::stream::Stream;
+
+use timely_logging::Logger;
+
+use crate::logging::{CommunicationEvent, CommunicationEventBuilder, MessageEvent, StateEvent};
 
 fn tcp_panic(context: &'static str, cause: io::Error) -> ! {
     // NOTE: some downstream crates sniff out "timely communication error:" from
@@ -37,10 +39,9 @@ pub fn recv_loop<S>(
 where
     S: Stream,
 {
+    let mut logger = logger.map(|logger| logger.into_typed::<CommunicationEvent>());
     // Log the receive thread's start.
-    if let Some(logger) = logger.as_ref() {
-        logger.log(CommunicationEvent::from(StateEvent { send: false, process, remote, start: true }));
-    }
+    logger.as_mut().map(|l| l.log(StateEvent { send: false, process, remote, start: true }));
 
     let mut targets: Vec<MergeQueue> = targets.into_iter().map(|x| x.recv().expect("Failed to receive MergeQueue")).collect();
 
@@ -88,9 +89,9 @@ where
             let bytes = buffer.extract(peeled_bytes);
 
             // Record message receipt.
-            if let Some(logger) = logger.as_ref() {
-                logger.log(CommunicationEvent::from(MessageEvent { is_send: false, header }));
-            }
+            logger.as_mut().map(|logger| {
+                logger.log(MessageEvent { is_send: false, header, });
+            });
 
             if header.length > 0 {
                 stageds[header.target - worker_offset].push(bytes);
@@ -117,9 +118,7 @@ where
     }
 
     // Log the receive thread's end.
-    if let Some(logger) = logger.as_ref() {
-        logger.log(CommunicationEvent::from(StateEvent { send: false, process, remote, start: false, }));
-    }
+    logger.as_mut().map(|l| l.log(StateEvent { send: false, process, remote, start: false, }));
 }
 
 /// Repeatedly sends messages into a TcpStream.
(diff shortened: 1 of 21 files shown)


communication/src/allocator/zero_copy/tcp.rs
```diff
@@ -35,12 +33,14 @@ pub fn recv_loop<S>(
...
     remote: usize,
-    mut logger: Option<Logger<CommunicationEvent>>)
+    logger: Option<Logger<CommunicationEventBuilder>>)
 where
     S: Stream,
 {
     // Log the receive thread's start.
-    logger.as_mut().map(|l| l.log(StateEvent { send: false, process, remote, start: true }));
+    if let Some(logger) = logger.as_ref() {
+        logger.log(CommunicationEvent::from(StateEvent { send: false, process, remote, start: true }));
```
frankmcsherry, line 42: I've seen a bunch of these, and don't know yet but wonder if `log` should take a `T: From<Foo>` where `Foo` is the type it otherwise expects as its argument.

antiguru: That's hard because we need a T that can be copied into the container (builder), and that supports From for some argument type. I don't think rust could figure out the in-between type because there might be more than one option. This means we have to specify it somewhere manually....

logging/src/lib.rs
```diff
@@ -133,7 +132,7 @@ impl<T> Logger<T> {
     /// This implementation borrows a shared (but thread-local) buffer of log events, to ensure
     /// that the `action` only sees one stream of events with increasing timestamps. This may
     /// have a cost that we don't entirely understand.
-    pub fn log<S: Into<T>>(&self, event: S) {
+    pub fn log<S>(&self, event: S) where CB: PushInto<(Duration, S)> {
```
frankmcsherry, line 135: Can you say why the `Into<T>` went away? It feels like a bit of the noise in other files is making up for this. Is it that we shouldn't want to invoke `.into()` before pushing, or .. some other reason? Or no particular reason?

logging/src/lib.rs
```diff
@@ -151,108 +150,69 @@ impl<T> Logger<T> {
...
 
 /// Flush on the *last* drop of a logger.
-impl<T, A: ?Sized + FnMut(&Duration, &mut Vec<(Duration, T)>)> Drop for LoggerInner<T, A> {
+impl<CB: ContainerBuilder, A: ?Sized + FnMut(&Duration, &mut CB::Container)> Drop for LoggerInner<CB, A> {
     fn drop(&mut self) {
         // Avoid sending out empty buffers just because of drops.
-        if !self.buffer.is_empty() {
-            self.flush();
-        }
+        self.flush();
```
frankmcsherry, line 190: The code doesn't really match the comment any more. I think it does avoid sending empty buffers because of `flush()`'s implementation, but also the implementation has a TODO to change that.

timely/src/dataflow/scopes/child.rs
```diff
@@ -70,6 +70,7 @@ where
     fn peek_identifier(&self) -> usize {
         self.parent.peek_identifier()
     }
+    #[inline(always)]
```
frankmcsherry, line 73: ?

timely/src/logging.rs
```diff
@@ -142,9 +150,9 @@ pub struct TimelyProgressEvent {
     /// Sequence of nested scope identifiers indicating the path from the root to this instance.
     pub addr: Vec<usize>,
     /// List of message updates, containing Target descriptor, timestamp as string, and delta.
-    pub messages: Box<dyn ProgressEventTimestampVec>,
+    pub messages: std::rc::Rc<dyn ProgressEventTimestampVec>,
```
frankmcsherry, line 153: You mentioned that this change was due to needing `Clone` to work out. I think we should probably at least write a note about this explaining to ourselves what is going on here. Another option (h/t @parkmycar)  is to have the traits provide a `box_clone(&self)` method that produces a `Box<dyn Trait>`, though .. we would need to write the `Clone` implementation manually. I'm not 100% sure I grok the fall out with `Rc`, if for example we ever needed to implement `Send`.

timely/src/progress/reachability.rs
```diff
@@ -74,7 +74,6 @@
 
 use std::collections::{BinaryHeap, HashMap, VecDeque};
 use std::cmp::Reverse;
-
```
frankmcsherry, line 77: Hrm. Not sure about the formatting, but generally try to space out `std` vs crate stuffs!

frankmcsherry (approved):
Although I think we can merge this, it seems like there is some noise around how we log things, both in the switch from `.map(|l| ...)` to `if let`, but also in removing the `Into` constraint requiring more awareness about the types that actually get logged. Happy to discuss!

communication/src/allocator/zero_copy/tcp.rs
frankmcsherry: Gotcha. One option would be to have a `Logger<T, CB: PushInto<T>>`, which carries the opinions about what you should be pushing in to it. I can't really tell whether it is worse giving up the generality of `PushInto`, as that would do, or giving up the `Into` constraint and asking folks to build the logged types more explicitly.

logging/src/lib.rs
```diff
@@ -151,108 +161,144 @@ impl<T> Logger<T> {
...
-            if self.buffer.len() == self.buffer.capacity() {
-                // Would call `self.flush()`, but for `RefCell` panic.
-                (self.action)(&elapsed, &mut self.buffer);
-                // The buffer clear could plausibly be removed, changing the semantics but allowing users
-                // to do in-place updates without forcing them to take ownership.
-                self.buffer.clear();
-                let buffer_capacity = self.buffer.capacity();
-                if buffer_capacity < Self::buffer_capacity() {
-                    self.buffer.reserve((buffer_capacity+1).next_power_of_two());
-                }
```
antiguru, line 200: We had logic in place that tried to grow the buffers geometrically, but I'm not sure that was still working as originally intended. With the switch to container builders, that logic certainly doesn't exist here anymore.

antiguru:
I changed the PR to introduce a `TypedLogger` that supports logging types that are `Into` for some other type that we know how to log. This allows us to remove most of the changes around how we log events.

Also, I followed the `box_clone` suggestion, which means that we don't need to switch to `Rc`s anymore, with the downside of additional `Clone` implementations.

I restored the flush behavior to send empty buffers when nothing had been logged, and drop to only flush if anything had been logged since the last flush.

frankmcsherry (approved):
Looks good! I think I understand most of it, but I haven't fully grokked the `dirty` stuff yet. Sorry if me being confused resulted in PR churn. The other parts look good, and we should move ahead with this, and I can figure out what we need to be true about logging async, though it almost certainly derives from running it and seeing what is bad, which landing this unblocks more of.

PR #619, opened 2025-01-12
Log action can distinguish data from flush

antiguru:
Previously, the log action would receive an empty container when we'd like
to report progress but no new data. We'd only send the empty container when
we didn't have any other data to send. This makes it hard for users to
distinguish between flushing and just data updates. It might be important
to distinguish the two because only on flush we might not be called again
for a while, but otherwise it's very likely that the logger might receive
more data, or sees a flush.

This change alters the signature of the action to accept a `&mut Option<C>`
(where `C` is `CB::Container`), and we pass `Some(container)` on data, and
`None` on flush.

Clients using the logging API need to change their implementation, as both
vectors and `Option` offer a `iter` function, but with different results.

### DIFF
--- a/logging/src/lib.rs
+++ b/logging/src/lib.rs
@@ -7,8 +7,9 @@ use std::time::{Instant, Duration};
 use std::fmt::{self, Debug};
 use std::marker::PhantomData;
 
-use timely_container::{ContainerBuilder, PushInto};
+use timely_container::{Container, ContainerBuilder, PushInto};
 
+/// A registry binding names to typed loggers.
 pub struct Registry {
     /// A map from names to typed loggers.
     map: HashMap<String, (Box<dyn Any>, Box<dyn Flush>)>,
@@ -28,7 +29,9 @@ impl Registry {
     /// seen (likely greater or equal to the timestamp of the last event). The end of a
     /// logging stream is indicated only by dropping the associated action, which can be
     /// accomplished with `remove` (or a call to insert, though this is not recommended).
-    pub fn insert<CB: ContainerBuilder, F: FnMut(&Duration, &mut CB::Container)+'static>(
+    ///
+    /// Passing a `&mut None` container to an action indicates a flush.
+    pub fn insert<CB: ContainerBuilder, F: FnMut(&Duration, &mut Option<CB::Container>)+'static>(
         &mut self,
         name: &str,
         action: F) -> Option<Box<dyn Any>>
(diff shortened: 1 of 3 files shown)


logging/src/lib.rs
```diff
@@ -234,57 +233,56 @@ impl<CB: ContainerBuilder, T> std::ops::Deref for TypedLogger<CB, T> {
...
         let elapsed = self.time.elapsed() + self.offset;
         for event in events {
-            self.dirty = true;
             self.builder.push_into((elapsed, event.into()));
             while let Some(container) = self.builder.extract() {
-                (self.action)(&elapsed, container);
+                let mut c = Some(std::mem::take(container));
+                (self.action)(&elapsed, &mut c);
+                if let Some(mut c) = c {
+                    c.clear();
```
frankmcsherry, line 247: No strong opinion, but is this clearing new behavior? Does it e.g. prevent passing back owned data, into which the logger can write? Again, really no strong opinion, but just checking whether the force-clear is new, and whether it is intentional.

logging/src/lib.rs
```diff
@@ -234,57 +233,56 @@ impl<CB: ContainerBuilder, T> std::ops::Deref for TypedLogger<CB, T> {
...
+            (self.action)(&elapsed, &mut c);
+            if let Some(mut c) = c {
+                c.clear();
+                *container = c;
+            }
         }
 
-        self.dirty = false;
+        // Send no container to indicate flush.
+        (self.action)(&elapsed, &mut None);
```
frankmcsherry, line 267: A nit, but one of the reasons to take a `&mut Option<_>` rather than a `Option<&mut _>` is to allow the `None` call to pass back resources, and at least with `Push` the intent is that you keep calling this as long as you get a non-`None` back. We don't have to do that here, and probably massively over-thinking this, but wanted to call out the gap.

frankmcsherry (approved):
This all looks good to me. I left a few comments, but they are roughly nits, and can be followed on later if we conclude they are worth it!

logging/src/lib.rs
antiguru: I see your point, but I'm wondering if this is actually true! Or put differently, is it something we'd _like_ to be true, or is it something that we had at some point and forgot about?

The reason I'm asking is because most (all?) places where we call `Push::done`, we don't have a loop to drain resources. If I recall correctly, the only place where we loop is in Differential merge batchers to drain the stash of allocations once we're done merging chains.

We could change `Push`'s `done` function to look like this instead:
```
    fn done(&mut self) { 
        let mut container = None;
        loop {
            self.push(&mut container);
            if container.is_none() { break; }
        }
    }
```

logging/src/lib.rs
antiguru: It was purely defensive. Thinking about it, it should be the container builder's responsibility to enforce what needs to be true about a container after extracting or finishing it, so it doesn't make sense to have the `clear` call here.

PR #621, opened 2025-01-15
give_container for arbitrary container builders

antiguru:
Previously, we'd only support `Session::give_container` for capacity container builders, but there is no reason not to offer it for all builders. Here, we change the definition to apply for all session and buffers. The implementation ensures FIFO order by flushing before pushing a whole prefabricated container.


frankmcsherry:
This looks good to me, thanks!

PR #624, opened 2025-01-17
Typed logging

antiguru:
Convert the progress and reachability logs to typed logs, which do not require casting at runtime.

The loggers are accessible using new names:
* `timely/progress` -> `timely/progress/` with the type name appended.
* `timely/reachability` -> `timely/reachability/` with the type name appended.

It's using `std::any::type_name()` to determine a unique string for the timestamp type.

Analyzing `logging-send` for allocations, this roughly cuts down total allocations in half.

`target/release/examples/logging-send 1000 1000 -w4` produces roughly 850k allocations in current master, and 445k on this branch.

### DIFF
--- a/timely/src/progress/broadcast.rs
+++ b/timely/src/progress/broadcast.rs
@@ -28,12 +28,12 @@ pub struct Progcaster<T:Timestamp> {
     /// Communication channel identifier
     channel_identifier: usize,
 
-    progress_logging: Option<ProgressLogger>,
+    progress_logging: Option<ProgressLogger<T>>,
 }
 
 impl<T:Timestamp+Send> Progcaster<T> {
     /// Creates a new `Progcaster` using a channel from the supplied worker.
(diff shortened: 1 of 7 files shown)


frankmcsherry (approved):
This looks pretty good to me; no notes.

timely/src/progress/broadcast.rs
```diff
@@ -134,8 +134,8 @@ impl<T:Timestamp+Send> Progcaster<T> {
             // options for improving it if performance limits users who want other logging.
             self.progress_logging.as_ref().map(|l| {
 
-                let mut messages = Box::new(Vec::with_capacity(changes.len()));
-                let mut internal = Box::new(Vec::with_capacity(changes.len()));
+                let mut messages = Vec::with_capacity(changes.len());
+                let mut internal = Vec::with_capacity(changes.len());
```
antiguru, line 138: Note that we should be using `recv_changes.len()` here, because `changes` doesn't have any relation to what we're receiving. But, then we'd over-allocate :/

frankmcsherry: Ah, could you explain that more? I agree that `recv_changes.len()` looks more appropriate, but what is the downside (seems like we fill these up with exactly that many elements, and then ship them in the `TimelyProgressEvent`)?

PR #626, opened 2025-01-17
Log operator summaries using clever approach

frankmcsherry:
Attempt at logging operator summaries (the internal matrix of input->output timestamp summaries describing the operator's internal connectivity). I mostly copied existing code, and .. once this is up I'll be able to have a better look and see if it makes any sense. It roughly follows #624 which did something similar for progress and reachability logging, though I may have erred in important ways.

### DIFF
--- a/timely/src/logging.rs
+++ b/timely/src/logging.rs
@@ -10,6 +10,10 @@ pub type TimelyLogger = crate::logging_core::TypedLogger<TimelyEventBuilder, Tim
 pub type TimelyProgressEventBuilder<T> = CapacityContainerBuilder<Vec<(Duration, TimelyProgressEvent<T>)>>;
 /// Logger for timely dataflow progress events (the "timely/progress" log stream).
 pub type TimelyProgressLogger<T> = crate::logging_core::Logger<TimelyProgressEventBuilder<T>>;
+/// Container builder for timely dataflow progress events.
+pub type TimelySummaryEventBuilder<TS> = CapacityContainerBuilder<Vec<(Duration, OperatesSummaryEvent<TS>)>>;
+/// Logger for timely dataflow progress events (the "timely/progress" log stream).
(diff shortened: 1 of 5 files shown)


timely/src/logging.rs
```diff
@@ -10,6 +10,10 @@ pub type TimelyLogger = crate::logging_core::TypedLogger<TimelyEventBuilder, Tim
 pub type TimelyProgressEventBuilder<T> = CapacityContainerBuilder<Vec<(Duration, TimelyProgressEvent<T>)>>;
 /// Logger for timely dataflow progress events (the "timely/progress" log stream).
 pub type TimelyProgressLogger<T> = crate::logging_core::Logger<TimelyProgressEventBuilder<T>>;
+/// Container builder for timely dataflow progress events.
+pub type TimelySummaryEventBuilder<TS> = CapacityContainerBuilder<Vec<(Duration, OperatesSummaryEvent<TS>)>>;
+/// Logger for timely dataflow progress events (the "timely/progress" log stream).
```
antiguru, line 15: Incorrect comments

timely/src/progress/subgraph.rs
```diff
@@ -636,7 +641,8 @@ impl<T: Timestamp> PerOperatorState<T> {
         mut scope: Box<dyn Operate<T>>,
         index: usize,
         identifier: usize,
-        logging: Option<Logger>
+        logging: Option<Logger>,
+        summary_logging: &mut Option<SummaryLogger<T::Summary>>,
```
antiguru, line 645: Alternatively `Option<&mut SummaryLogger<...>>`? I can't seem to ever know which one to use where, but here we don't want to change the option... Also, logging mightn't need mutable access iirc

antiguru (approved):
Looks great, I left some comments!

timely/src/progress/subgraph.rs
frankmcsherry: Will address in a follow-up!

PR #628, opened 2025-01-22
Flatten reachability logging, log identifier

antiguru:
Previously, reachability and progress logging would use their address to identify the place where an event happened. In logging this is not great because every event needs to come with a vector. Instead, use the operator's global identifier, which we can translate back to the address using the operates events.

### DIFF
--- a/timely/src/dataflow/scopes/mod.rs
+++ b/timely/src/dataflow/scopes/mod.rs
@@ -57,15 +57,6 @@ pub trait Scope: ScopeParent {
     /// an operator will be added, and it is an error not to eventually add such an operator.
     fn allocate_operator_index(&mut self) -> usize;
 
-    /// Adds a child `Operate` to the builder's scope using a supplied index.
-    ///
-    /// This is used internally when there is a gap between allocate a child identifier and adding the
-    /// child, as happens in subgraph creation.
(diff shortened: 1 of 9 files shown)


timely/src/dataflow/scopes/mod.rs
```diff
@@ -57,15 +57,6 @@ pub trait Scope: ScopeParent {
...
 
-    /// Adds a child `Operate` to the builder's scope using a supplied index.
-    ///
-    /// This is used internally when there is a gap between allocate a child identifier and adding the
-    /// child, as happens in subgraph creation.
-    fn add_operator_with_index(&mut self, operator: Box<dyn Operate<Self::Timestamp>>, index: usize) {
-        let global = self.new_identifier();
-        self.add_operator_with_indices(operator, index, global);
-    }
-
```
frankmcsherry, line 68: Is there an explainer for this deletion, and pivoting the inputs to the `_indices` version? Seems like we only need to pivot the subgraph builder to use `_indices`; is this (and the input) an unrelated tidy?

antiguru: Remove the change from the PR.

frankmcsherry (approved):
This looks wonderful, and I appreciate the accommodation!

PR #629, opened 2025-01-22
Move opinions about reachability logging into `TrackerLogger`

frankmcsherry:
This is a very meh PR that moves some opinions about reachability logging into the `TrackerLogger`, allowing the user of the logger (the `Tracker`) to just hand over a consolidated iterator over (references to) updates. If we want to improve the `TrackerLogger` to e.g. use `columnar` which would not require cloning the times, that is a little bit easier now.

### DIFF
--- a/timely/src/progress/reachability.rs
+++ b/timely/src/progress/reachability.rs
@@ -573,22 +573,16 @@ impl<T:Timestamp> Tracker<T> {
             let target_changes =
             self.target_changes
                 .iter()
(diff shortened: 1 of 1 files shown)


antiguru (approved):
I think we can merge the change as it moves some of the behavior specific to logging closer to the code that logs reachability events, which is a good thing. At the same time, I'm not fully convinced this is what we'll need to, say, support columnar, but that shouldn't stand in the way!

PR #632, opened 2025-02-01
capability: avoid touching `ChangeBatch` on no-op downgrades

petrosagg:


frankmcsherry:
My bias here is not to merge, but let me explain the thinking and you can let me know if I am missing something. The test makes sense, but it is logic that the caller could always perform. The work the test introduces is not optional for callers that do not need it (e.g. those maintaining mutable antichains and reacting only to changes). So, baking in the additional test that is overhead for some, and convenience for others, is .. semi-antithetical to how the crate is set up.

If the `PartialOrder` methods were ternary, and you could see `Equal` and `Less`, then it feels like an easy merge.

petrosagg:
> So, baking in the additional test that is overhead for some, and convenience for others, is .. semi-antithetical to how the crate is set up.

Ah, yes I agree that this test doesn't fit within that frame of thought. Thanks for the explanation!

PR #635, opened 2025-02-10
Update columnar to 0.3, and columnar example

frankmcsherry:
Columnar updated with a breaking change to serialization, and this adjusts how the columnar example uses that serialization. The good news is that it should be better now, because there is a new serialization format (though we could continue to use the old one), and the way one accesses the format is easier to access reliably (vs the hard coded serialization in the example).


antiguru (approved):
Seems good! Thank you

PR #637, opened 2025-02-11
bytes: Separate `Bytes` and `BytesMut`

frankmcsherry:
This PR separates `Bytes` into `Bytes` and `BytesMut`. The former implements `clone()` and is unwriteable. The latter does not implement clone but is writeable. One goes from the latter to the former with the `freeze()` method.

The intent is to minimize the use of `BytesMut`, which is narrowed only to moments within the writing to a `BytesSlab`. As soon as the writing completes it yields a `Bytes` which is from that point on unwriteable.

One of the goals is to make serialized messages cloneable, so that broadcast channels can more efficiently be implemented.

### DIFF
--- a/bytes/src/lib.rs
+++ b/bytes/src/lib.rs
@@ -96,7 +96,7 @@ pub mod arc {
                 sequestered: self.sequestered.clone(),
             };
 
-            unsafe { self.ptr = self.ptr.add(index); }
+            self.ptr = self.ptr.wrapping_add(index);
             self.len -= index;
 
             result
@@ -161,7 +161,7 @@ pub mod arc {
         /// shared4.try_merge(shared2).ok().expect("Failed to merge 4 and 231");
         /// ```
         pub fn try_merge(&mut self, other: Bytes) -> Result<(), Bytes> {
-            if Arc::ptr_eq(&self.sequestered, &other.sequestered) && ::std::ptr::eq(unsafe { self.ptr.add(self.len) }, other.ptr) {
+            if Arc::ptr_eq(&self.sequestered, &other.sequestered) && ::std::ptr::eq(self.ptr.wrapping_add(self.len), other.ptr) {
                 self.len += other.len;
                 Ok(())
             }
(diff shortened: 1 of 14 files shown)


frankmcsherry:
The changes to `bytes/src/lib.rs` are a bit silly to read as a diff. As the functionality was split between the two, the diff inserts a type definition halfway through an impl block. It can also be read as "copy `Bytes` to `BytesMut`; make `Bytes.ptr` a const ptr; delete methods that do not get used".

bytes/src/lib.rs
```diff
@@ -138,6 +139,77 @@ pub mod arc {
             }
         }
 
+        /// Converts a writeable byte slice to an shareable byte slice.
```
antiguru, line 142:
```suggestion
        /// Converts a writeable byte slice to a shareable byte slice.
```

bytes/src/lib.rs
```diff
@@ -51,15 +52,15 @@ pub mod arc {
         sequestered: Arc<dyn Any>,
     }
 
-    // Synchronization happens through `self.sequestered`, which mean to ensure that even
+    // Synchronization happens through `self.sequestered`, which means to ensure that even
     // across multiple threads each region of the slice is uniquely "owned", if not in the
     // traditional Rust sense.
-    unsafe impl Send for Bytes { }
+    unsafe impl Send for BytesMut { }
```
antiguru, line 58: Consider removing the `Send` implementation for `BytesMut`. Timely compiles and tests without it.

antiguru (approved):
I think this looks good. I left some comments inline, but the overall structure makes sense to me.

You might want to add `#[inline]` annotations to simple functions (`freeze`, `deref*`) so we don't have to go back in the future should we notice that they're not inlined. As they don't have generic parameters, without LTO, they wouldn't be inlined.

bytes/src/lib.rs
frankmcsherry: Great point! I also added some notes about things that are not implemented but could be, if we wanted a more broadly usable crate (for the moment, no such goal).

PR #641, opened 2025-02-13
Demonstrate `BytesSlab` with abstracted bytes storage

frankmcsherry:
This PR demonstrates how one could use a `BytesSlab` without specifying the `T: DerefMut<Target=[u8]>` and instead using a `Box<dyn ...>` instead.


frankmcsherry:
Merging, but I think the next step will be to tidy up the config options, so that e.g. `refill` doesn't end up as an argument to `new_vector`, but rather a member of a config that supports a `new_vector(&self)` method.

PR #642, opened 2025-02-13
Zero-copy for networked processes

antiguru:
Introduce a intra- and inter-process serializing configuration variant.
Previously, it wasn't possible to construct a networked configuration that
would use zero-copy for process-local communication.


frankmcsherry:
Looks great, and thanks!

PR #647, opened 2025-02-18
Remove columnation and flatcontainer deps

antiguru:
Remove the dependencies on columnation and flatcontainer as far as
possible. This removes the TimelyStack type because it's not used in Timely
and only retains the columnation implementation for Product. Flatcontainer
was only used as an example, and we don't have any downstream dependencies
on it left.

Support for columnation and flatcontainer can be added by downstream crates
if they chose to do so.


frankmcsherry:
Looks good, and quite a tidy; thank you!

PR #651, opened 2025-03-15
Linear connectivity

frankmcsherry:
The PR updates how we express operator-internal connectivity information, from `input` and `output` indexed dense vectors `Vec<Vec<Antichain<TS>>>`, to `Vec<BTreeMap<usize, Antichain<TS>>>`. This change reduces the complexity from `inputs * outputs` to `inputs + edges`, which is a substantial reduction for operators with many inputs and outputs, and relatively sparse connectivity.

The PR introduces two type definitions, which could become `struct` types in the own right:
```rust
/// Operator internal connectivity, from inputs to outputs.
pub type Connectivity<TS> = Vec<PortConnectivity<TS>>;
/// Internal connectivity from one port to any number of opposing ports.
pub type PortConnectivity<TS> = std::collections::BTreeMap<usize, Antichain<TS>>;
```

As it turns out, we only perform random access into the `BTreeMap` structures as part of set-up, so the performance penalty for random access is bounded. Otherwise, we iterate over the contents when moving progress information forward, which should be roughly as performant with `BTreeMap::iter()` as our previous iteration and enumeration.

A `shape_scaling.rs` test is added, for both operators with many inputs and outputs, and subgraphs with many inputs and outputs. The test goes up to 100k of each; 1M does complete, in roughly linear time, but it exceeds 60s. Memory requirements are not small (hundreds of megabytes), but seem to be linear.

### DIFF
--- a/timely/src/dataflow/operators/generic/builder_rc.rs
+++ b/timely/src/dataflow/operators/generic/builder_rc.rs
@@ -20,7 +20,7 @@ use crate::dataflow::operators::capability::Capability;
 use crate::dataflow::operators::generic::handles::{InputHandleCore, new_input_handle, OutputWrapper};
 use crate::dataflow::operators::generic::operator_info::OperatorInfo;
 use crate::dataflow::operators::generic::builder_raw::OperatorShape;
-
+use crate::progress::operate::PortConnectivity;
 use crate::logging::TimelyLogger as Logger;
 
 use super::builder_raw::OperatorBuilder as OperatorBuilderRaw;
@@ -33,7 +33,7 @@ pub struct OperatorBuilder<G: Scope> {
     consumed: Vec<Rc<RefCell<ChangeBatch<G::Timestamp>>>>,
     internal: Rc<RefCell<Vec<Rc<RefCell<ChangeBatch<G::Timestamp>>>>>>,
     /// For each input, a shared list of summaries to each output.
-    summaries: Vec<Rc<RefCell<Vec<Antichain<<G::Timestamp as Timestamp>::Summary>>>>>,
+    summaries: Vec<Rc<RefCell<PortConnectivity<<G::Timestamp as Timestamp>::Summary>>>>,
     produced: Vec<Rc<RefCell<ChangeBatch<G::Timestamp>>>>,
     logging: Option<Logger>,
 }
@@ -64,7 +64,7 @@ impl<G: Scope> OperatorBuilder<G> {
     where
         P: ParallelizationContract<G::Timestamp, C> {
 
-        let connection = (0..self.builder.shape().outputs()).map(|_| Antichain::from_elem(Default::default())).collect();
+        let connection = (0..self.builder.shape().outputs()).map(|o| (o, Antichain::from_elem(Default::default()))).collect();
         self.new_input_connection(stream, pact, connection)
     }
 
@@ -76,7 +76,7 @@ impl<G: Scope> OperatorBuilder<G> {
     ///
     /// Commonly the connections are either the unit summary, indicating the same timestamp might be produced as output, or an empty
     /// antichain indicating that there is no connection from the input to the output.
-    pub fn new_input_connection<C: Container, P>(&mut self, stream: &StreamCore<G, C>, pact: P, connection: Vec<Antichain<<G::Timestamp as Timestamp>::Summary>>) -> InputHandleCore<G::Timestamp, C, P::Puller>
+    pub fn new_input_connection<C: Container, P>(&mut self, stream: &StreamCore<G, C>, pact: P, connection: PortConnectivity<<G::Timestamp as Timestamp>::Summary>) -> InputHandleCore<G::Timestamp, C, P::Puller>
         where
             P: ParallelizationContract<G::Timestamp, C> {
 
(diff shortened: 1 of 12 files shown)


timely/src/dataflow/operators/generic/builder_rc.rs
```diff
@@ -108,7 +108,7 @@ impl<G: Scope> OperatorBuilder<G> {
     /// antichain indicating that there is no connection from the input to the output.
     pub fn new_output_connection<CB: ContainerBuilder>(
         &mut self,
-        connection: Vec<Antichain<<G::Timestamp as Timestamp>::Summary>>
+        connection: PortConnectivity<<G::Timestamp as Timestamp>::Summary>
```
antiguru, line 111: This could be an iterator over `(input, summary)`, with the caveat that duplicate `input` elements will override previous mentions.

timely/src/progress/reachability.rs
```diff
@@ -156,11 +157,15 @@ impl<T: Timestamp> Builder<T> {
...
+    pub fn add_node(&mut self, index: usize, inputs: usize, outputs: usize, summary: Connectivity<T::Summary>) {
 
         // Assert that all summaries exist.
         debug_assert_eq!(inputs, summary.len());
-        for x in summary.iter() { debug_assert_eq!(outputs, x.len()); }
+        for x in summary.iter() { 
+            for o in x.keys() {
+                debug_assert!(o < &outputs);
+            }
+        }
```
antiguru, line 168: I wonder if Rust is smart enough to eliminate the loop if the body is empty. I haven't seen the old version of the loop ever stand out, so probably not an issue?

timely/src/progress/reachability.rs
```diff
@@ -749,7 +754,20 @@ fn summarize_outputs<T: Timestamp>(
...
 
-    let mut results: HashMap<Location, Vec<Antichain<T::Summary>>> = HashMap::new();
+    // A reverse map from operator outputs to inputs, along their internal summaries.
+    let mut reverse_internal = HashMap::new();
+    for (node, connectivity) in nodes.iter().enumerate() {
+        for (input, outputs) in connectivity.iter().enumerate() {
+            for (output, summary) in outputs.iter() {
+                reverse_internal
+                    .entry(Location::new_source(node, *output))
+                    .or_insert_with(|| Vec::default())
```
antiguru, line 764:
```suggestion
                    .or_default()
```

timely/src/progress/reachability.rs
```diff
@@ -771,25 +789,21 @@ fn summarize_outputs<T: Timestamp>(
...
-                                worklist.push_back((location, output, combined));
+            Port::Source(_output_port) => {
+                if let Some(inputs) = reverse_internal.get(&location) {
+                    for (input_port, operator_summary) in inputs.iter() {
+                        let location = Location::new_target(location.node, *input_port);
+                        for summary in operator_summary.elements().iter() {
+                            if let Some(combined) = summary.followed_by(&summary) {
+                                let update =
+                                results.entry(location)
+                                    .or_insert_with(|| PortConnectivity::default())
```
antiguru, line 800:
```suggestion
                                    .or_default()
```

timely/src/progress/reachability.rs
```diff
@@ -771,25 +789,21 @@ fn summarize_outputs<T: Timestamp>(
...
+                                    .entry(output)
+                                    .or_insert_with(|| Antichain::default())
```
antiguru, line 802:
```suggestion
                                    .or_default()
```

timely/src/progress/reachability.rs
```diff
@@ -803,14 +817,9 @@ fn summarize_outputs<T: Timestamp>(
...
                 if let Some(&source) = reverse.get(&location) {
-                    let antichains = results
-                        .entry(source)
-                        .and_modify(|antichains| antichains.reserve(output))
-                        .or_insert_with(|| Vec::with_capacity(output));
-
-                    while antichains.len() <= output { antichains.push(Antichain::new()); }
-
-                    if antichains[output].insert(summary.clone()) {
+                    let antichains = results.entry(source).or_insert_with(|| PortConnectivity::default());
```
antiguru, line 820:
```suggestion
                    let antichains = results.entry(source).or_default();
```

timely/src/progress/reachability.rs
```diff
@@ -803,14 +817,9 @@ fn summarize_outputs<T: Timestamp>(
...
+                    let update = antichains.entry(output).or_insert_with(|| Antichain::default()).insert_ref(&summary);
```
antiguru, line 821:
```suggestion
                    let update = antichains.entry(output).or_default().insert_ref(&summary);
```

timely/src/progress/subgraph.rs
```diff
@@ -555,13 +555,15 @@ where
...
         // with each element containing `self.outputs()` antichains regardless
         // of how long `self.scope_summary` is
-        let mut internal_summary = vec![vec![Antichain::new(); self.outputs()]; self.inputs()];
-        for (input_idx, input) in self.scope_summary.iter().enumerate() {
-            for (output_idx, output) in input.iter().enumerate() {
-                let antichain = &mut internal_summary[input_idx][output_idx];
-                antichain.reserve(output.elements().len());
-                antichain.extend(output.elements().iter().cloned().map(TInner::summarize));
-            }
+        // let mut internal_summary: Connectivity<TOuter::Summary> = vec![PortConnectivity::default(); self.inputs()];
```
antiguru, line 558: Stray comment

antiguru (approved):
Some nits, but looks good otherwise. Thank you!

timely/src/dataflow/operators/generic/builder_rc.rs
frankmcsherry: Yeah, I thought of this, and .. figured I'd wait before being too clever. Mostly wanted the highest confidence diff first. But you are right, and it's probably tasteful to do (especially given that the `new_output` method conses up one just to have it be dropped).

timely/src/progress/reachability.rs
frankmcsherry: I'd guess it does, but that's a good point that this could be written differently (and is, elsewhere in the code). Brb!

timely/src/dataflow/operators/generic/builder_rc.rs
frankmcsherry: Ah, punting for the moment to think a bit more. The argument is used twice, once passed to the raw builder, which has a similar method that could have a similar change. I'd like to get my head around why everything is duplicated first.

PR #653, opened 2025-03-19
Replace crossbeam-channel by the std variant

antiguru:
Replaces crossbeam-channel with the channels in the standard library. The implementations in the standard library are derived from crossbeam, but seem to be more actively maintained.


frankmcsherry:
This is independently appealing, in terms of reducing the scope of dependencies, but does this derive from a more conclusive look at the ASAN double-frees?

antiguru:
I'd hope the code in `std` gets tested more than other crates, and the experience Materialize had with 0.5.14, I'd say we switch to the standard library implementation. Also removes one dependency.

Issue #654, opened 2025-03-19
Potentially corrupt connectivity after linear connectivity changes


antiguru:
When pointing Materialize and Differential at Timely with #651, I observe crashes in `validate_progress`:

```
Increment at (1742294913660, PointStamp { vector: [1] }), not supported by
        consumed: [ChangeBatch { updates: [], clean: 0 }, ChangeBatch { updates: [], clean: 0 }]
        internal: MutableAntichain { updates: ChangeBatch { updates: [((1742294913660, PointStamp { vector: [2] }), 1), ((1742390934535, PointStamp { vector: [] }), 1)], clean: 2 }, frontier: [(1742294913660, PointStamp { vector: [2] }), (1742390934535, PointStamp { vector: [] })], changes: ChangeBatch { updates: [], clean: 0 } }
2025-03-19T13:28:55.277366Z  thread 'timely:work-0' panicked at /home/moritz/.cargo/git/checkouts/timely-dataflow-dbb5eb8e114d5497/d0ea86f/timely/src/progress/subgraph.rs:790:25:
Progress error; internal "BuildingObject(System(535))"
   7: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   8: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   9: <timely::progress::subgraph::PerOperatorState<timely::order::product::Product<mz_repr::timestamp::Timestamp, differential_dataflow::dynamic::pointstamp::PointStamp<u64>>>>::validate_progress
             at /home/moritz/.cargo/git/checkouts/timely-dataflow-dbb5eb8e114d5497/d0ea86f/timely/src/progress/subgraph.rs:790:25
  10: <timely::progress::subgraph::Subgraph<mz_repr::timestamp::Timestamp, timely::order::product::Product<mz_repr::timestamp::Timestamp, differential_dataflow::dynamic::pointstamp::PointStamp<u64>>>>::activate_child
             at /home/moritz/.cargo/git/checkouts/timely-dataflow-dbb5eb8e114d5497/d0ea86f/timely/src/progress/subgraph.rs:373:17
[...]

PR #656, opened 2025-03-24
Revert "Linear connectivity (#651)"

antiguru:
This reverts commit d0ea86fb6a8c540c9775908250ab34a62be7ee87.

We've observed reachability problems with #651, so revert it until fixed. See #654 for context.


frankmcsherry:
Will re-attempt with a more measured approach, swapping in new logic w/o swapping out old logic yet.

PR #658, opened 2025-03-24
Linear reachability, take 2

frankmcsherry:
Slowly building up more confidence in changing our internal operator reachability implementation. This is a more careful version of #651, which in particular avoids a shadowing bug that seemed to be at the root of #654.

The external facing changes are that customer operator input/output creation no longer take a `Vec<Antichain<_>>`, but rather an `I: IntoIterator<Item=(usize, Antichain<_>)>`. This allows a sparse description of connectivity, in cases where it would be expensive to produce a large number of `Antichain::default()` empty antichains.

The new behavior is scale-tested in `tests/shape_scaling.rs`, where we test operators and subgraphs up to 100,000 x 100,000.

### DIFF
--- a/timely/src/progress/operate.rs
+++ b/timely/src/progress/operate.rs
@@ -61,8 +61,63 @@ pub trait Operate<T: Timestamp> : Schedule {
 /// Operator internal connectivity, from inputs to outputs.
 pub type Connectivity<TS> = Vec<PortConnectivity<TS>>;
 /// Internal connectivity from one port to any number of opposing ports.
-pub type PortConnectivity<TS> = Vec<Antichain<TS>>;
+#[derive(serde::Serialize, serde::Deserialize, columnar::Columnar, Debug, Clone, Eq, PartialEq)]
+pub struct PortConnectivity<TS> {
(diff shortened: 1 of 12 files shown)


timely/src/progress/operate.rs
```diff
@@ -58,6 +58,55 @@ pub trait Operate<T: Timestamp> : Schedule {
...
+
+impl<TS> Default for PortConnectivity<TS> {
+    fn default() -> Self {
+        Self { tree: std::collections::BTreeMap::new() }
+    }
+}
+
+impl<TS> PortConnectivity<TS> {
+    /// Inserts an element by reference, ensuring that the index exists.
+    pub fn insert(&mut self, index: usize, element: TS) -> bool where TS : crate::PartialOrder + Clone {
```
antiguru, line 77:
```suggestion
    pub fn insert(&mut self, index: usize, element: TS) -> bool where TS : crate::PartialOrder {
```
Nit: We don't need `Clone` here.

antiguru (approved):
Looks good; left one inline comment.

PR #663, opened 2025-05-07
Make `std::time::Instant` optional

frankmcsherry:
There are a few uses of `std::time::Instant` in the codebase which do not *need* to be there. This is the first commit from https://github.com/TimelyDataflow/timely-dataflow/pull/577, and something we can discuss independent of the follow-on commits that rework the scheduler.


antiguru (approved):
Looks good, thank you!

PR #668, opened 2025-06-16
Optionally flush container builders

antiguru:
Flush container builders in exchange channels and session buffers, if they need to. This avoids a problem where we'd have a quadratic amount of lingering memory in the exchange channel. The change adds a `relax` function to container builders, which by default does nothing. Container builders can use it to release resources when called, and Timely calls it when it believes it's a good moment to flush. This corresponds to pushing a `None` value, which happens once an operator ceases to produce data.

I don't like the solution very much because it seems to conflate different concepts, but it seems like a small API change without too much negative impact.

Edit: changed the name from `flush` to `relax` for the lack of a better term.

### DIFF
--- a/timely/examples/columnar.rs
+++ b/timely/examples/columnar.rs
@@ -335,6 +335,10 @@ mod builder {
 
         #[inline]
         fn relax(&mut self) {
+            /// The caller is responsible for draining all contents; assert that we are empty.
+            /// The assertion is not strictly necessary, but it helps catch bugs.
+            debug_assert!(self.current.is_empty());
+            debug_assert!(self.pending.is_empty());
             *self = Self::default();
         }
     }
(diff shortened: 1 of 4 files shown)


timely/examples/columnar.rs
```diff
@@ -331,6 +332,11 @@ mod builder {
             self.empty = self.pending.pop_front();
             self.empty.as_mut()
         }
+
+        #[inline]
+        fn relax(&mut self) {
+            *self = Self::default();
+        }
```
frankmcsherry, line 339: Should we maybe `debug_assert` that this is empty? That's the intent, right? It should be a semantic no-op, but potentially release resources.

frankmcsherry (approved):
Seems good in spirit, but have a few nits about the API. Mostly being clear exactly what `relax` should do, and making sure it does this in the one implementation.

PR #671, opened 2025-06-19
Log channel container type

antiguru:
Logs the channel container type as part of the `ChannelsEvent`. This allows us to surface the channel container type in debug data.


frankmcsherry:
Seems cheap, only on channel creation. Great!

PR #676, opened 2025-06-25
Use broadcast channel for the broadcast operator

antiguru:
Timely recently gained a broadcast channel. This change switches the broadcast operator to using the broadcast channel. It also tests that Arc can be broadcasted.

This requires us to enable the "rc" feature on serde.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,7 @@ edition = "2021"
 
 [workspace.dependencies]
 columnar = "0.6"
+serde = { version = "1.0", features = ["derive", "rc"] }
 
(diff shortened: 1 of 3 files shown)


antiguru:
Unfortunately this is not correct. In its current form it'd confuse progress tracking as each count on the input appears worker-times on the output. It's not a simple fix to change this, as we count before handing data off to a `Tee`. The pusher only sees what the `Tee` hands it, but at that time the counts are fixed already. Also, it doesn't have access to the produced counts.

Closing the PR because more work is necessary to enable this functionality.

PR #677, opened 2025-06-25
Test mdbook using regular doc tests

antiguru:


frankmcsherry:
Let's see! (( I don't understand everything, but up for trying it! ))

PR #680, opened 2025-07-05
Update for Columnar 0.7

frankmcsherry:
Changes required to target `columnar`'s changes in 0.7. Combined with a few clean-ups, `examples/columnar.rs` gets a fair bit nicer in my opinion!

Columnar 0.7 isn't live yet, so this should fail until it is released.

### DIFF
--- a/timely/examples/columnar.rs
+++ b/timely/examples/columnar.rs
@@ -1,4 +1,4 @@
-//! Wordcount based on flatcontainer.
+//! Wordcount based on the `columnar` crate.
 
 use {
     std::collections::HashMap,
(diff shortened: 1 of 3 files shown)


timely/examples/columnar.rs
```diff
@@ -19,7 +19,8 @@ struct WordCount {
 
 fn main() {
 
-    type Container = Column<WordCount>;
+    type InnerContainer = <WordCount as columnar::Columnar>::Container;
```
antiguru, line 22: We could have a typedef in columnar:
```
pub type ContainerOf<T> = <T as Columnar>::Container;
```

PR #686, opened 2025-07-21
Progress container with length

antiguru:
Splits the Container trait into a progress container and a container. The
progress container only exposes a length, which is enough for most parts of
Timely. The container trait extends the progress container and carries the
usual functions to iterate and clear the contents.

I'm open to renaming and moving types, and primarily wanted to check
whether it's possible to split the trait.

Signed-off-by: Moritz Hoffmann <moritz@turmeric.pc.hksc.de>


container/src/lib.rs
```diff
@@ -4,46 +4,34 @@
...
-        self.len() == 0
+        self.count() == 0
     }
+}
 
-    /// Remove all contents from `self` while retaining allocated memory.
-    /// After calling `clear`, `is_empty` must return `true` and `len` 0.
-    fn clear(&mut self);
+/// A container that can reveal its contents through iterating by reference and draining.
+pub trait IterableContainer: WithProgress {
```
frankmcsherry, line 29: I foresee this being deleted / unified with `columnar`'s container traits. No rush, but curious if that checks out too. This is now I think not timely specific as much as "can I read data out of here?". At least, if not columnar then some random container traits somewhere, unrelated to core timely, in timely only as long as there aren't better ones somewhere else.

container/src/lib.rs
```diff
@@ -4,46 +4,34 @@
...
-    type Item<'a> where Self: 'a;
-
-    /// Push `item` into self
-    #[inline]
-    fn push<T>(&mut self, item: T) where Self: PushInto<T> {
-        self.push_into(item)
-    }
-
+/// We require [`Default`] for convenience purposes.
+pub trait WithProgress: Default {
```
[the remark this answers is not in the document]
frankmcsherry: Tbh, I kinda like `Container` as the name here, in that from timely's point of view containers hold numbers of things that should be tracked, and it has no other opinion. The other trait being `ItemContainer` or `IterContainer` or whathaveyou makes lots of sense, but partly because it isn't as "core timely" function.

container/src/lib.rs
```diff
@@ -102,7 +90,7 @@ pub trait PushInto<T> {
 /// decide to represent a push order for `extract` and `finish`, or not.
 pub trait ContainerBuilder: Default + 'static {
     /// The container type we're building.
-    type Container: Container + Clone + 'static;
+    type Container: WithProgress + Default + Clone + 'static;
```
frankmcsherry, line 93: `WithProgress : Default`, so is the `+ Default` redundant, or perhaps helpful because that `WithProgress` constraint isn't locked down?

container/src/lib.rs
```diff
@@ -140,7 +128,34 @@ pub trait ContainerBuilder: Default + 'static {
 /// Specifically, the sum of lengths of all extracted and finished containers must equal the
 /// number of times that `push_into` is called on the container builder.
 /// If you have any questions about this trait you are best off not implementing it.
-pub trait LengthPreservingContainerBuilder : ContainerBuilder { }
+pub trait CountPreservingContainerBuilder: ContainerBuilder { }
```
frankmcsherry, line 131: Names are a bit weird here, because "count" is a method on `WithProgress` and not obviously a top-level concept. I wonder if it makes sense to require all container builders to be count preserving? As in, you now need to implement both `count()` and `len()`, and there's no room for containers whose `count()` method is incorrect?

timely/src/dataflow/channels/mod.rs
```diff
@@ -51,7 +52,6 @@ impl<T, C: Container> Message<T, C> {
 
         if let Some(message) = bundle {
             *buffer = message.data;
-            buffer.clear();
```
frankmcsherry, line 54: This line prompted me to think I don't actually understand what is going on. It seems like a pretty substantial change to some contracts, and it's worth going through the reason for it.

timely/src/dataflow/channels/pushers/buffer.rs
```diff
@@ -44,7 +44,7 @@ impl<T, CB: Default, P> Buffer<T, CB, P> {
     }
 }
 
-impl<T, C: Container + Data, P: Push<Message<T, C>>> Buffer<T, CapacityContainerBuilder<C>, P> where T: Eq+Clone {
+impl<T, C: SizableContainer + Data, P: Push<Message<T, C>>> Buffer<T, CapacityContainerBuilder<C>, P> where T: Eq+Clone {
```
frankmcsherry, line 47: This looks like a new constraint, e.g. as opposed to `IterableContainer`. Is there a reason?

frankmcsherry (commented):
Reading through and I think I've got some comments recorded to start with, before continuing with what seems like a lot of search-replace rewrites. Thoughts at the moment are that
1. This is directionally a good idea, but:
2. We should take this opportunity to boil away some complexity. Do we still need traits like "length preserving" if we have a concept of a container that must report its count accurately and everyone needs to implement that?

I have some biases around naming that we can dance around. I'm currently leaning towards `Container` for `WithProgress`: everything we ship around timely should have this information, and just having that be what "container" means works great for me. The bonus features around "you can enumerate some unrelated number of records from this container" are helpful, but not as core and I'm up for letting their names reflect this.

container/src/lib.rs
```diff
@@ -4,46 +4,34 @@
...
-    /// imposed on the `LengthPreservingContainerBuilder` trait, whose implementors
+    /// imposed on the [`CountPreservingContainerBuilder`] trait, whose implementors
     /// must preserve the number of items.
-    fn len(&self) -> usize;
+    fn count(&self) -> usize;
 
-    /// Determine if the container contains any elements, corresponding to `len() == 0`.
+    /// Determine if the container contains any elements, corresponding to `count() == 0`.
+    #[inline(always)]
     fn is_empty(&self) -> bool {
```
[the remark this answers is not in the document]
antiguru: Checked the places where we use this, and it is mostly to guard against sending empty containers. Is it correct to send empty containers, i.e., should the count for something within progress tracking be non-zero? It doesn't make much sense to me to send containers with length 0, so it's probably good to guard against it.

container/src/lib.rs
```diff
@@ -4,46 +4,32 @@
...
-    type ItemRef<'a> where Self: 'a;
-
-    /// The type of elements when draining the container.
-    type Item<'a> where Self: 'a;
-
-    /// Push `item` into self
-    #[inline]
-    fn push<T>(&mut self, item: T) where Self: PushInto<T> {
-        self.push_into(item)
-    }
```
antiguru, line 29: `push` wasn't used, removed.

container/src/lib.rs
```diff
@@ -4,46 +4,34 @@
 
 use std::collections::VecDeque;
 
-/// A container transferring data through dataflow edges
+/// A type representing progress, with an update count.
```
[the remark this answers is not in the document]
antiguru: Interesting. We currently have the separation into `Message` and the data it's transferring, and the `Container` trait in the past was just the data, mostly for historic reasons. It might make sense to make _messages_ the interface to sending data, which would give more control about the associated timestamps to the users.

container/src/lib.rs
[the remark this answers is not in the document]
antiguru: Renamed to Container!

container/src/lib.rs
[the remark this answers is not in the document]
antiguru: (Removed `is_empty` and replaced all calls with `.count() > 0`).

container/src/lib.rs
antiguru: Removed, was redundant.

container/src/lib.rs
antiguru: Wasn't the idea of the trait to indicate that it's compatible with `partition`, i.e., after partitioning, the sum length of the parts equals the length of the input? From this perspective, it seems worth keeping around.

I restored the old name.

timely/src/dataflow/channels/mod.rs
frankmcsherry: Right. Elsewhere (e.g. `columnar`) you just have a `Clear` trait and require it when you want to be able to re-use a container.

timely/src/dataflow/channels/pushers/buffer.rs
frankmcsherry: I'm confused then. `SizableContainer` is weaker than `Container`? Surely `SizableContainer : Container`?

antiguru: Turns out we don't need this change. Removed.

container/src/lib.rs
antiguru: It might still be useful for a container builder that compacts in-place, like we have in Differential. The consolidating container builder certainly can only be used before we fix counts for progress tracking. It can't be used for an exchange because data might consolidate differently.

I see two ways out: Either we have a container that stores the count independently of the contents, and sets the count to the number of `push_into` calls _outside_ progress tracking, or rework the exchange logic to make it clear that container builders inside have less flexibility than the ones before we take the count.

For this reason, I think it makes sense to leave this trait in Timely for the time being.

PR #689, opened 2025-08-12
Downgrade capability validity test to `debug_assert`

frankmcsherry:
The capability validity test for input capabilities of output capabilities is linear in the number of outputs, which means that to create sessions for all outputs is quadratic in the number of outputs.

The better fix here, which we can investigate, is to support better bulk session creation, or for validity checking to come with a hint about where to look for the corresponding summary. These are a bit more invasive, but one of them probably makes sense to do.


antiguru (approved):
Thank you!

PR #691, opened 2025-08-12
Avoid capability validity scan with explicit index

frankmcsherry:
Timely's capability validity test would use a scan to find any pair of port and buffer that matched, but we actually know the intended port information and can communicate it directly, as a hint, to avoid the scan. At the moment the test is only in a `debug_assert!`, but this would make it easier to return that test to an `assert!`.

### DIFF
--- a/timely/src/dataflow/operators/capability.rs
+++ b/timely/src/dataflow/operators/capability.rs
@@ -250,7 +250,7 @@ impl<T: Timestamp> CapabilityTrait<T> for InputCapability<T> {
         let internal_borrow = self.internal.borrow();
         // To be valid, the output buffer must match and the timestamp summary needs to be the default.
         Rc::ptr_eq(&internal_borrow[port], query_buffer) &&
-        summaries_borrow.get(port).map(|path| path.len() == 1 && path[0] == Default::default()).unwrap_or(false)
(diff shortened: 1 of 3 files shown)


timely/src/dataflow/operators/capability.rs
```diff
@@ -245,14 +245,12 @@ pub struct InputCapability<T: Timestamp> {
...
+    fn valid_for_output(&self, query_buffer: &Rc<RefCell<ChangeBatch<T>>>, port: usize) -> bool {
         let summaries_borrow = self.summaries.borrow();
         let internal_borrow = self.internal.borrow();
         // To be valid, the output buffer must match and the timestamp summary needs to be the default.
-        let result = summaries_borrow.iter_ports().any(|(port, path)| {
-            Rc::ptr_eq(&internal_borrow[port], query_buffer) && path.len() == 1 && path[0] == Default::default()
-        });
-        result
+        Rc::ptr_eq(&internal_borrow[port], query_buffer) &&
+        summaries_borrow.get(port).map(|path| path.len() == 1 && path[0] == Default::default()).unwrap_or(false)
```
antiguru, line 253:
```suggestion
        summaries_borrow.get(port).map_or(false, |path| path.elements() == [Default::default()])
```

antiguru (approved):
Seems good!

PR #692, opened 2025-08-13
partition: lazily create and then reuse output sessions

petrosagg:
The partition operator was previously unconditionally initializing sessions for all outputs every time it received an incoming container even though the container might have contained datums for only a small subset of the outputs.

This PR makes it so sessions are only lazily initialized when the first datum for them is observed. Additionally, sessions are reused across incoming containers as long as the session's capability dominates the incoming ones.


frankmcsherry (approved):
I understand this now; thanks!

PR #695, opened 2025-08-14
capture: use length prefixed encoding to reliably receive events

petrosagg:
The existing capture/replay mechamism would produce a concatenation of bincode encoded messages. However, on the receiving end the bincode format does not have enough information for messages to be decoded accurately from a non-delimited stream, which could lead to the receiving end getting confused.

This PR fixes the issue by prefixing each messages with its length in bytes.

### DIFF
--- a/timely/src/dataflow/operators/core/capture/capture.rs
+++ b/timely/src/dataflow/operators/core/capture/capture.rs
@@ -29,7 +29,7 @@ pub trait Capture<T: Timestamp, C: Container + Data> {
     /// use std::rc::Rc;
     /// use std::sync::{Arc, Mutex};
     /// use timely::dataflow::Scope;
-    /// use timely::dataflow::operators::{Capture, ToStream, Inspect};
+    /// use timely::dataflow::operators::{Capture, ToStream};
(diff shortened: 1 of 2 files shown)


timely/src/dataflow/operators/core/capture/capture.rs
```diff
@@ -79,30 +79,26 @@ pub trait Capture<T: Timestamp, C: Container + Data> {
...
-    ///     recv.set_nonblocking(true).unwrap();
+    /// // these allow us to capture / replay a timely stream.
+    /// let list = TcpListener::bind("127.0.0.1:8001").unwrap();
+    /// let send = TcpStream::connect("127.0.0.1:8001").unwrap();
+    /// let recv = list.incoming().next().unwrap().unwrap();
+    /// recv.set_nonblocking(true).unwrap();
     ///
-    ///     worker.dataflow::<u64,_,_>(|scope1|
+    /// std::thread::scope(move |s| {
+    ///     s.spawn(move || timely::example(move |scope1| {
```
petrosagg, line 89: I moved the two sides in separate threads since if one operator blocks while writing to the socket nothing will unblock it

frankmcsherry (approved):
Looks good, and tyvm!

PR #697, opened 2025-08-16
Rework container abstractions

antiguru:
This is another attempt at reworking the container abstractions in Timely.

At the moment, we have two types that a lot of the container abstractions hinge on, `Container` and `ContainerBuilder`. Both do things that are of _localized_ interest, among the things they need to do. Containers can explain their contents, can clear themselves, even though it's not a primary concern of a container. What it absolutely needs to do is to explain to progress tracking how many updates it contains. The container builder can partition itself, but only if the container is special, i.e., it provides internal access.

This PR contains individual commits to do the following, roughly ordered by the sequence of commits:
* Remove `Container::push` because it wasn't used and would just defer to `push_into`.
* Remove `Container::clear`. This is somewhat problematic as forgetting to clear can cause resource leaks or repeating data that we already processed earlier. The good news is that we're using container builders in most places, and container builders need to carefully recycle the containers they produced. They're also in a much better position to clear because they know what kind of container they're handling. I looked through all the call sites in Timely, and convinced myself this change is safe.
* Split the container trait into `Container`, `IterContainer`, `DrainContainer`, without depending on each other. The container provides a length for progress tracking, the iter and drain containers allow to iterate or drain the contents. It's a bit strange as the drain container doesn't _need_ to be empty after draining, probably something to look at later.
* Remove some unneeded `is_empty` calls where we know containers are non-empty.
* Rename (and change type) of `Container::len` to `update_count`. Document it as specific to progress tracking.
* Rename `Container` to `WithProgress` because that's what it's about.
* Introduce a `Partitioner` trait in exchange that knows how to partition a specific container into pushers. Remove the `partition` function from the container builder. This change is probably worth looking at individually, or even as a separate PR.

None of this is intended as "this is the only way it has to happen", but rather a sequence of steps that might get us to a clearer definition of the crucial types we use in Timely.

The location of the traits is up to discussion, and maybe we conclude we don't need the container crate anymore and rather localize the traits.


timely/src/dataflow/operators/core/capture/extract.rs
```diff
@@ -48,10 +48,10 @@ pub trait Extract<T, C> {
     fn extract(self) -> Vec<(T, C)>;
 }
 
-impl<T: Ord, C: SizableContainer> Extract<T, C> for ::std::sync::mpsc::Receiver<Event<T, C>>
+impl<T, C> Extract<T, C> for ::std::sync::mpsc::Receiver<Event<T, C>>
 where
-    for<'a> C: PushInto<C::Item<'a>>,
-    for<'a> C::Item<'a>: Ord,
+    for<'a> C: SizableContainer + DrainContainer<Item<'a>: Ord> + PushInto<C::Item<'a>>,
```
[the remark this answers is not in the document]
antiguru: Incorrect suggestion.

container/src/lib.rs
```diff
@@ -4,63 +4,55 @@
...
-    }
-
-    /// The number of elements in this container
+/// It must implement default for historic reason. The default implementation is not required
+/// to allocate memory for variable-length components.
+// TODO: Remove `Default` requirement in the future.
+// The container is `Default` because `CapacityContainerBuilder` only implements `ContainerBuilder`
+// for containers that implement `Default`, and we use the associated `::Container` all over Timely.
+// We can only access the type if all requirements for the `ContainerBuilder` implementation are
+// satisfied.
```
antiguru, line 19: Remove comment

container/src/lib.rs
```diff
@@ -4,63 +4,55 @@
...
+    /// of outstanding updates, and it is highly load bearing. The main restriction is
     /// imposed on the `LengthPreservingContainerBuilder` trait, whose implementors
     /// must preserve the number of items.
-    fn len(&self) -> usize;
-
-    /// Determine if the container contains any elements, corresponding to `len() == 0`.
-    fn is_empty(&self) -> bool {
-        self.len() == 0
-    }
+    fn update_count(&self) -> i64;
```
antiguru, line 27: Rename to record_count

container/src/lib.rs
```diff
@@ -4,63 +4,55 @@
...
+    /// The type of elements when reading non-destructively from the container.
+    type ItemRef<'a> where Self: 'a;
     /// Iterator type when reading from the container.
     type Iter<'a>: Iterator<Item=Self::ItemRef<'a>> where Self: 'a;
-
     /// Returns an iterator that reads the contents of this container.
     fn iter(&self) -> Self::Iter<'_>;
+}
 
+/// TODO
```
antiguru, line 43: Add documentation

container/src/lib.rs
```diff
@@ -137,7 +138,7 @@ pub trait ContainerBuilder: Default + 'static {
 
 /// A wrapper trait indicating that the container building will preserve the number of records.
 ///
-/// Specifically, the sum of lengths of all extracted and finished containers must equal the
+/// Specifically, the sum of record counts of all extracted and finished containers must equal the
 /// number of times that `push_into` is called on the container builder.
```
frankmcsherry, line 142: Nit, but .. I can imagine `PushInto` being implemented for iterators and other things that might deposit multiple records. I'm not sure we exclude that, and we might want the text to either be more ambiguous on this matter (i.e. not say it must be the number of pushes), or less ambiguous about the right thing (.. not really sure what to say here, awkwardly :D).

I think we want everything that is pushed to have an implied "number of records", and the count should match the accumulation of those numbers, but we don't actually require anything about the `PushInto` implementation. Which is fine (at least, complicated to codify), but maybe some alternate text here would help. As an example:
```
/// Specifically, the sum of record counts of all extracted and finished containers must equal the number of accounted records that are pushed into the container builder
```

timely/src/dataflow/operators/core/filter.rs
```diff
@@ -23,16 +23,14 @@ pub trait Filter<C: Container> {
...
 where
     for<'a> C: PushInto<C::Item<'a>>
 {
     fn filter<P: FnMut(&C::Item<'_>)->bool+'static>(&self, mut predicate: P) -> StreamCore<G, C> {
         self.unary(Pipeline, "Filter", move |_,_| move |input, output| {
             input.for_each(|time, data| {
-                if !data.is_empty() {
-                    output.session(&time).give_iterator(data.drain().filter(&mut predicate));
-                }
+                output.session(&time).give_iterator(data.drain().filter(&mut predicate));
```
frankmcsherry, line 33: I was about to leave a comment about how this is a potential regression, and then concluded that I don't really care. :D

PR #702, opened 2025-08-27
Retire target and source changes in bulk

antiguru:
If we have many updates per source or target, `propagate_all` will update
an operator's pointstamps as many times as there are updates. Updating
the mutable antichain behind pointstamps can be expensive.

This change extracts the updates from the target and source changes, and
applies them in bulk on the pointstamps in the hope of reducing the cost
of maintaining the antichain.

### DIFF
--- a/timely/src/progress/change_batch.rs
+++ b/timely/src/progress/change_batch.rs
@@ -56,6 +56,24 @@ impl<T, const X: usize> ChangeBatch<T, X> {
         }
     }
 
+    /// Constructs a `ChangeBatch` from updates.
+    ///
+    /// # Examples
+    ///
+    ///```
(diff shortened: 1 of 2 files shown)


frankmcsherry:
Ah, so I understand what is going on here and it seems like a good idea. In particular, there can be a quadratic behavior here if an antichain for some `target` contains many times each with maybe a count of one, and then receives a batch of updates that retract each of those times. I bet we could test that theory with a pretty simple rig (but also can just believe it is true).

I was hoping to find something around `Peekable` that would allow a `take_while` adapter, perhaps something like https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.peeking_take_while.

PR #707, opened 2025-08-29
Remove IterContainer

antiguru:
This change removes the IterContainer type and reworks other types and
implementations to function independently. Notably, this is a breaking
change for users of exotic containers:
* The `Inspect` trait defers to `&C: IntoIterator` to reveal items in
  containers. Not all currently used containers provide this. Vec does,
  but Rc/Arc and Column don't and do not plan to in the future. Users
  should pivot to `inspect_container` instead and internalize the iteration
  in the closure.
* The `DrainContainer` implementaiton for Rc/Arc depends on the wrapped
  type implementing `IntoIterator` for references. This limits where
  wrapped containers that do not provide this can be used. It's mostly
  limited to the core Timely layer that doesn't provide element-by-element
  functionality.

I feel this change is net positive as it defers to the Rust API to iterate
existing types, instead of us providing our own infrastructure. It comes
at a cost of potentially breaking existing code, which seems acceptable.

### DIFF
--- a/container/src/lib.rs
+++ b/container/src/lib.rs
@@ -4,7 +4,7 @@
 
 use std::collections::VecDeque;
 
-/// An type containing a number of records accounted for by progress tracking.
+/// A type containing a number of records accounted for by progress tracking.
 ///
 /// The object stores a number of updates and thus is able to describe it count
 /// (`update_count()`) and whether it is empty (`is_empty()`). It is empty if the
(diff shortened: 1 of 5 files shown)


frankmcsherry (approved):
I'm good with all of this! Great, even! Moves away from timely as a row-by-row system, and aims at at moving containers of collections around, and if you want the row-by-row interpretation cool but that's some amount of on you.

PR #709, opened 2025-09-16
Create pact for Distributors

frankmcsherry:
This PR introduces the `DistributorPact<B>` which wraps a `B: FnOnce(usize) -> impl Distributor<C>`. Or, at least when it does, then it implements `ParallelizationContract`, which is a fast path from writing a distributor, writing an appropriate `usize` based constructor, and getting out a parallelization contract.

I also organized `pact.rs` a bit with modules; I hope it is cleaner now. And I added a default empty implementation to `Distributor::relax` which I hope is ok.


frankmcsherry:
Confirmed that this works with `columnar.rs` in differential, and the main outcome was deleting a bunch of code and imports that this logic handles instead.

PR #714, opened 2025-09-17
Remove `Container` argument from `Counter`

frankmcsherry:
The `Counter` push wrapper had a container as an argument, but it seems to be extraneous. It shows up instead first as part of the `Push` implementation, but not as a part of the type itself.

### DIFF
--- a/timely/src/dataflow/channels/pact.rs
+++ b/timely/src/dataflow/channels/pact.rs
@@ -110,11 +110,7 @@ mod distributor {
         C: Accountable + ContainerBytes + Send + 'static,
         D: Distributor<C> + 'static,
     {
-        type Pusher = Exchange<
-            T,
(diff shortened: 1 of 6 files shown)


timely/src/dataflow/channels/pact.rs
```diff
@@ -110,11 +110,7 @@ mod distributor {
         C: Accountable + ContainerBytes + Send + 'static,
         D: Distributor<C> + 'static,
     {
-        type Pusher = Exchange<
-            T,
-            LogPusher<Box<dyn Push<Message<T, C>>>>,
-            D
-        >;
+        type Pusher = Exchange<T, LogPusher<Box<dyn Push<Message<T, C>>>>, D>;
```
frankmcsherry, line 113: Unrelated, sorry. But seems great! :D

antiguru (approved):
Looks good!

PR #715, opened 2025-09-18
Pivot container builders above capability checking

frankmcsherry:
Timely has historically had a built in `pushers::Buffer` which houses a container builder, and is essentially mandatory. This is confusing for operators that do not require container builders, and from an abstraction point of view, the choice of how to build containers seems at a higher level than e.g. capability verification.

This PR pivots these around, introducing a `pushers::Progress` that validates capabilities, around which we wrap container building abstractions.

One regression that we should consider testing is that `.session(&cap)` returns a `Session` which will flush the container builder when dropped, which is unlike our current behavior where .. it doesn't flush anything and if you create a new session with the same time it just picks up where it left off. I think that is a glitch, and creates weird logical states where you have an `Option<T>` in some longer-lived state. We could reintroduce it though, by adding the `Option` to the `OutputBuffer`.


frankmcsherry:
There's a thing I'd like to fix, which is the ability to group `InputHandle::next` output into something that iterates through `(time, iter)` where `iter` is some iterator over containers. This would give a reasonable answer to the session chaining that we lost with this PR, but would actually be a "better" way to do it. Pretty small change, but interested to talk through this before doing too much about it. It should be able to do done on main rather than in the PR, for example.

timely/src/dataflow/operators/core/partition.rs
```diff
@@ -46,43 +46,41 @@ impl<G: Scope, C: Container + DrainContainer> Partition<G, C> for StreamCore<G,
...
-                            }
-                        };
-                        session.give(datum2);
+                    while let Some((part, data)) = parts.pop_first() {
+                        for datum in data.into_iter() {
+                            c_build.push_into(datum);
+                        }
+                        while let Some(container) = c_build.finish() {
+                            handles[part as usize].give(&cap, container);
+                        }
```
antiguru, line 83: Might be better at reusing allocations if written as:
```suggestion
                        for datum in data.into_iter() {
                            c_build.push_into(datum);
                            while let Some(container) = c_build.finish() {
                                handles[part as usize].give(&cap, container);
                            }
                        }
```

timely/src/dataflow/operators/core/capture/replay.rs
```diff
@@ -118,8 +117,9 @@ where
                     activator.activate_after(delay);
                 }
 
-                output.cease();
-                output.inner().produced().borrow_mut().drain_into(&mut progress.produceds[0]);
+                use timely_communication::Push;
+                output.push(&mut None);
```
antiguru, line 121: I think we should either push `&mut None` and remove `Push::done`, or use `Push::done` consistently. No big deal, but removing a function might not be a bad call.

antiguru: Similarly, `Push` has a `send` function that we rarely use. Consider removing it.

timely/src/dataflow/operators/core/partition.rs
frankmcsherry: I think this would need to be `extract` rather than `finish`, and then a `finish` loop at the end? But I see what you are saying.

timely/src/dataflow/operators/core/capture/replay.rs
frankmcsherry: A related thing: there were moments where one needs to know `C: Container` earlier than you might want, on account of we need a `Push<_>` implementation to be able to call `done()`, even though `C` is not involved. Potentially a `Done` trait that provides the method, and then `Push: Done` or something. Maybe overly complicated, but removes at least one generic parameter from a type that doesn't seem to need it. Will investigate!

antiguru: I wouldn't mind a `Done` trait to avoid spreading the types from drop to the type and everywhere from there. I also think that we should keep the stack of pushers for exactly the reason that we might _not_ want to force a tee on everyone.

antiguru: It's been a while: https://github.com/TimelyDataflow/timely-dataflow/pull/508

antiguru (approved):
I'm good with merging this, and prefer addressing the downstream changes in one go rather than piece-by-piece.

PR #716, opened 2025-09-20
Add and use `InputSession`

frankmcsherry:
This PR re-imagines how we provide access to inputs to users, through an `InputSession` that can (and must) enumerate all `(time, data)` pairs. This encourages folks to use inputs idiomatically, rather than potentially draining inputs or potentially not (with "not" being a logical error because the operator may not be scheduled again).

The `InputSession` has one method that groups all `(time, data)` by `time` and provides `data` as an iterator over mutable container references. The migration path is pretty standard, but occasionally fiddly. Each `while let Some((time, data)) = input.next()` loop turns in to `input.for_each(|time, data| {` where `data` is one additional layer of iterator, and needs to be `flat_map`d or the like. This method consumes the input session, which addresses the `#[must_use]` attribute.

One alternative that could make the migration that much easier is to leave `for_each` as the container-at-a-time iteration that it is right now, and add a `by_time` or `for_time` or something that does the grouping. I think we want to encourage the latter, and making the migration lightly disruptive in order to do so might be fine.

### DIFF
--- a/timely/src/dataflow/operators/core/reclock.rs
+++ b/timely/src/dataflow/operators/core/reclock.rs
@@ -56,14 +56,14 @@ impl<S: Scope, C: Container> Reclock<S> for StreamCore<S, C> {
         self.binary_notify(clock, Pipeline, Pipeline, "Reclock", vec![], move |input1, input2, output, notificator| {
 
             // stash each data input with its timestamp.
-            input1.for_each(|cap, data| {
+            input1.for_each_time(|cap, data| {
                 for data in data {
                     stash.push((cap.time().clone(), std::mem::take(data)));
                 }
             });
 
             // request notification at time, to flush stash.
-            input2.for_each(|time, _data| {
+            input2.for_each_time(|time, _data| {
                 notificator.notify_at(time.retain());
             });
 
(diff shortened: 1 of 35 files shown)


timely/src/dataflow/operators/core/reclock.rs
```diff
@@ -57,7 +57,9 @@ impl<S: Scope, C: Container> Reclock<S> for StreamCore<S, C> {
 
             // stash each data input with its timestamp.
             input1.for_each(|cap, data| {
-                stash.push((cap.time().clone(), std::mem::take(data)));
+                for data in data {
+                    stash.push((cap.time().clone(), std::mem::take(data)));
+                }
```
frankmcsherry, line 62: This is a moment where the un-grouped iterator could have helped the migration, as we seem to want a capability per container.

frankmcsherry:
One candidate change is to have `for_each(|time, data| ...)` go over pairs of capability and container, but be "deprecated" or whatever it takes to whine at users, and a new `by_time` perform the grouping we do here. Mostly thinking it would be nice to ensure a smooth path for existing users, and I think this old `for_each` ensures that, even though the newer one is what we'd like folks to use.

frankmcsherry:
Merging to move forward, but lmk in review if there are things to fix up, or if we should revert for some reason! :D

PR #783, opened 2026-04-10
Remove Clone from Scope; OperatorBuilder takes &mut Scope

antiguru:
## Summary

- Remove `Clone` impl from `Scope<T>` to prevent accidental scope escape (which would cause `Rc::try_unwrap` to fail when finalising the subgraph)
- `OperatorBuilder::new()` takes `&mut Scope<T>` and extracts the inner `Rc<RefCell<SubgraphBuilder<T>>>` + `Worker` parts — no lifetime on the builder
- `Stream<T, C>` stores subgraph Rc + Worker directly instead of a `Scope`; removes `scope()` method
- `OperatorSlot<T>` stores subgraph Rc instead of a full `Scope`; removes `scope_mut()`
- `scoped`/`region`/`iterative` closures now receive the parent scope as the first argument: `scope.region(|outer, inner| { ... .leave(outer) })`, since the parent can no longer be captured via clone
- Remove unused `progress_logging` field from `Scope`
- Add `progress_log_name`/`summary_log_name`/`reachability_log_name` helpers in `logging.rs`
- Fix flaky capture doc test (hardcoded port 8000 → port 0)
- Add logging tests for progress and summary events (flat + nested/Product)

## Test plan

- [x] `cargo check` clean (no warnings)
- [x] `cargo test` passes (154/154, the only prior failure was the fixed port 8000 test)
- [x] New logging tests cover progress and summary events with content assertions
- [ ] Review API ergonomics of the new closure signatures

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mdbook/src/chapter_2/chapter_2_4.md
+++ b/mdbook/src/chapter_2/chapter_2_4.md
@@ -79,15 +79,19 @@ extern crate timely;
 
 use timely::dataflow::operators::Inspect;
 use timely::dataflow::operators::generic::operator::source;
+use timely::scheduling::Scheduler;
 
 fn main() {
     timely::example(|scope| {
 
+        // Acquire activations before borrowing scope mutably.
+        let activations = scope.activations();
+
         source(scope, "Source", |capability, info| {
 
             // Acquire a re-activator for this operator.
-            use timely::scheduling::Scheduler;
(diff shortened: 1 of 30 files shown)


antiguru:
We shouldn't do this!

PR #789, opened 2026-04-15
Spilling MergeQueue

frankmcsherry:
Summary by Claude:

  Three traits in zero_copy::spill:                                                       
                                                                                          
  - SpillPolicy — fn apply(&mut self, queue: &mut VecDeque<QueueEntry>). Rewrites queue   
  entries in place. Used on both sides: the writer's policy pages data out (Bytes → 
  Paged), the reader's policy pages data back in (Paged → Bytes).                         
  - BytesSpill — fn spill(&mut self, chunks: &mut Vec<Bytes>, handles: &mut Vec<Box<dyn 
  BytesFetch>>). Drains from chunks, pushes to handles. May stop partway on failure; both
  vecs are left in a consistent state.
  - BytesFetch — fn fetch(self: Box<Self>) -> Result<Vec<Bytes>, Box<dyn BytesFetch>>.
  Consumes the handle; returns data or returns itself for retry.

  One type alias:

  - SpillPolicyFn — Arc<dyn Fn() -> (Box<dyn SpillPolicy>, Box<dyn SpillPolicy>) + Send +
  Sync>. Factory producing matched (writer, reader) policy pairs, one per MergeQueue.

  Two shipped policies:

  - Threshold (writer) — spills entries past a head reserve when resident bytes exceed
  head_reserve_bytes + threshold_bytes.
  - PrefetchPolicy (reader) — materializes Paged entries near the front up to a byte
  budget.

  MergeQueue changes:

  - new_pair(buzzer, writer_policy, reader_policy) -> (MergeQueue, MergeQueue) —
  constructs a matched writer/reader pair. Each carries its own policy. Replaces
  new_spilling and build_reader.
  - extend runs the writer's policy after pushing.
  - drain_into runs the reader's policy before draining. Only pops Bytes entries; Paged
  entries stay for the next call. Buzzes if it returns empty with data still queued,
  preventing consumer stalls.
  - QueueEntry — Bytes(Bytes) | Paged(Box<dyn BytesFetch>). Internal to the queue;
  drain_into still yields Vec<Bytes> to callers.
  - Not Clone. new + new_pair are the constructors.

  Configuration via Hooks:

  - Hooks { refill, spill, log_fn } — bundles comm knobs previously passed as positional
  args. Config::try_build_with(hooks) consumes them. try_build() uses Hooks::default() (no
   spill). The spill field is Option<SpillPolicyFn>, threaded to every MergeQueue
  construction site (send-side, recv-side, intra-process).

  No new runtime dependencies (tempfile is dev-only). The file-backed strategy lives in
  the spill_stress example as a reference implementation.

### DIFF
--- a/communication/src/allocator/zero_copy/bytes_exchange.rs
+++ b/communication/src/allocator/zero_copy/bytes_exchange.rs
@@ -5,32 +5,36 @@ use std::collections::VecDeque;
 
 use timely_bytes::arc::Bytes;
 use super::bytes_slab::{BytesRefill, BytesSlab};
+use super::spill::{BytesFetch, SpillPolicy};
+
+/// An entry in a `MergeQueue`. Either `Bytes` resident in memory, or a
+/// handle to bytes previously written out via a `SpillPolicy`.
+pub enum QueueEntry {
+    /// Bytes resident in memory, ready to be consumed directly.
+    Bytes(Bytes),
+    /// Bytes spilled to a backing store, fetched via the handle.
+    Paged(Box<dyn BytesFetch>),
+}
 
 /// A target for `Bytes`.
 pub trait BytesPush {
-    // /// Pushes bytes at the instance.
-    // fn push(&mut self, bytes: Bytes);
     /// Pushes many bytes at the instance.
     fn extend<I: IntoIterator<Item=Bytes>>(&mut self, iter: I);
 }
 /// A source for `Bytes`.
 pub trait BytesPull {
-    // /// Pulls bytes from the instance.
-    // fn pull(&mut self) -> Option<Bytes>;
     /// Drains many bytes from the instance.
     fn drain_into(&mut self, vec: &mut Vec<Bytes>);
 }
 
 use std::sync::atomic::{AtomicBool, Ordering};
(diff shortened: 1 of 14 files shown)


communication/src/allocator/zero_copy/bytes_exchange.rs
```diff
@@ -102,7 +152,19 @@ impl BytesPull for MergeQueue {
         }
         let mut queue = lock_ok.expect("MergeQueue mutex poisoned.");
 
-        vec.extend(queue.drain(..));
+        // Drain queue entries under the lock, then release the lock before
+        // materializing any `Paged` entries via `fetch`, which may do I/O.
+        let drained: Vec<QueueEntry> = queue.drain(..).collect();
```
frankmcsherry: One option here is to drain into `vec` until we find a paged entry, and put entries from that point onward into the staging area.

frankmcsherry:
Caveat: I haven't read `spill.rs` other than the traits, and the "starter implementations" may be bonkers. Potentially right in spirit, but the main goal is the abstraction that allows one to rock up with their own implementation. I'll try and sort them out more tomorrow.

PR #792, opened 2026-05-12
Move ToStreamBuilder's container builder to method generic

antiguru:
Generic-over-`Item` trait with method-level `CB` enables method-call syntax `(0..3).to_stream_with_builder::<_, CB>(scope)` instead of the UFCS form `ToStreamBuilder::<CB>::to_stream_with_builder(...)`.

Breaking change to a public trait signature; CHANGELOG updated. Version bump handled separately.

### DIFF
--- a/timely/src/dataflow/operators/core/to_stream.rs
+++ b/timely/src/dataflow/operators/core/to_stream.rs
@@ -7,7 +7,7 @@ use crate::dataflow::operators::generic::operator::source;
 use crate::dataflow::{Stream, Scope};
 
 /// Converts to a timely [Stream], using a container builder.
-pub trait ToStreamBuilder<CB: ContainerBuilder> {
+pub trait ToStreamBuilder<Item> {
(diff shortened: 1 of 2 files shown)


timely/src/dataflow/operators/core/to_stream.rs
```diff
@@ -7,7 +7,7 @@ use crate::dataflow::operators::generic::operator::source;
 use crate::dataflow::{Stream, Scope};
 
 /// Converts to a timely [Stream], using a container builder.
-pub trait ToStreamBuilder<CB: ContainerBuilder> {
+pub trait ToStreamBuilder<Item> {
```
frankmcsherry: Minor feedback from the robot: this is perhaps better as an associated type, in that for the implementors there is only one type this could be. Would be potentially the only extension trait with an associated type, and .. perhaps this doesn't make it any more pleasant to type.

PR #793, opened 2026-05-15
Box pipeline pact behind dyn Push/Pull

antiguru:
Add `Allocate::pipeline` default returning `(Box<dyn Push<T>>, Box<dyn Pull<T>>)`, dispatched through `Process::pipeline`, `Allocator::pipeline`, and `Worker::pipeline`.
Hides `ThreadPusher` / `ThreadPuller` from callers and lets each allocator override the pipeline constructor.

A/B benchmark on `event_driven 10 100 record -w1` (10s wall, 10 trials each): boxed ~256k rounds, concrete ~255k rounds. Vtable indirection invisible — `LogPusher::push` does not show up in the top 25 functions in a samply profile.

### DIFF
--- a/communication/src/allocator/mod.rs
+++ b/communication/src/allocator/mod.rs
@@ -2,6 +2,7 @@
 
 use std::rc::Rc;
 use std::cell::RefCell;
+use std::sync::Arc;
 use std::time::Duration;
 
 pub use self::thread::Thread;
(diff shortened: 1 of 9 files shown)


antiguru:
## Performance

A/B comparison: this branch (`A`, boxed) vs `upstream/master` (`B`, concrete `ThreadPusher`/`ThreadPuller`).
Workload: `event_driven 10 100 record -w1` (10 dataflows × 100 map ops, Pipeline-heavy), 10s wall per trial, 10 trials each.
Apple M-series, arm64, rustc release profile.

Rounds completed in 10s (higher = better):

| build | A boxed | B concrete |
|---|---|---|
| default release | ~256k | ~255k |
| `lto = true`    | ~253k | ~254k |

Delta inside run-to-run noise (~1%) in both build modes.

### Asm

`LogPusher<Box<dyn Push<…>>>::push`: 152 bytes, ends with vtable load + indirect call (`ldp x8, x9, [x0]; ldr x9, [x9, #0x18]; blr x9`).
`LogPusher<ThreadPusher<…>>::push`: 548 bytes, full push chain inlined (CountPusher event count, VecDeque push, etc.), no indirect call.

### Profile (samply, `unionfind 200000 5000000 1 -w1`, ~5s)

`LogPusher::push` does not appear in the top 25 hot functions in either build.
Biggest deltas between the two profiles live in `MutableAntichain::rebuild` / `ChangeBatch::compact`, downstream of unrelated codegen choices triggered by the type change — not the indirection itself.

Vtable target is monomorphic per call site, so the indirect branch is BTB-predicted after warmup.
At ~1M Pipeline pushes per run, the ~2 extra instructions per push are ~2M cycles out of ~5G — below measurement noise.

communication/src/allocator/mod.rs
```diff
@@ -92,6 +105,40 @@ pub(crate) trait Allocate {
...
         (Box::new(Broadcaster { spare: None, pushers }), pull)
     }
+
+    /// Returns the runtime-supplied pipeline factory, if any.
+    ///
+    /// The default implementation returns `None`. Allocators that carry a
+    /// [`PipelineFactoryFn`] should override this to expose it; the default
+    /// [`Allocate::pipeline`] body consults it before falling back to the
+    /// thread-local channel.
+    fn pipeline_factory(&self) -> Option<&PipelineFactoryFn> { None }
```
antiguru, line 115: This function is a bit annoying, but the `Allocate` _trait_ doesn't have access to any state in the `pipeline` function.

antiguru:
Architecturally better to solve on higher levels on the stack; closing. If an operator wants to limit the amount of data pending on a pipeline channel, it can always retain some data and schedule itself again.

PR #797, opened 2026-05-22
frontier: hoist rebuild check out of update_iter

antiguru:
Exploration: factor the per-update rebuild check in `MutableAntichain::update_iter` into a `requires_rebuild` helper generic only over `T`, and replace the two `frontier.iter().any(...)` calls with a single explicit `for` loop. Deduplicates inlined `slice::Iter::any` bodies across `update_iter` monomorphizations and drops the per-call-site closure types. On the `event_driven` example, total LLVM lines drop by ~1600.


frankmcsherry:
Reads well to me; thank you!

PR #801, opened 2026-06-06
docs(mdbook): minor typo fixes

mike-vondy:
Was reading through the `mdbook` to get a grasp of the systems and noticed a small typo.

Typo `thought` to `though` in Chapter 2: Creating Inputs.

> In our first case, we've said we have an input but we haven't provided any clues about the type of data, and must do so to compile the program even though~~t~~ we don't care for the example.

Edit: Found another typo as I was going along so I figured I'd include it.
Added an `ly` to fix `relative` in `mdbook/src/chapter_4/chapter_4_4.md`

> The `capture_into` method is relative***ly*** simple, and we can just look at it:


frankmcsherry:
Thanks very much for the sharp eyes!

